PR #33, opened 2025-11-05
added local model support, and minor fixes to work on ubuntu

mrmartin:
added local model support, and tested with vllm running qwen3

It works and the results are similar to the reported ones:
Screenshot from 2025-11-05 15-13-53


rahulmoorthy19:
Thanks for putting up this integration code here!
Which parameter model of Qwen3 were you able to reproduce the results with?

Thanks!

PR #35, opened 2025-11-08
Use dependency-groups.dev

ifsheldon:
Hi! I got this warning when I used uv
```
warning: The `tool.uv.dev-dependencies` field (used in `pyproject.toml`) is deprecated and will be removed in a future release; use `dependency-groups.dev` instead
```

So, I just made this change to prevent future breakage due to the deprecation.


RobertTLange:
LGTM!

PR #41, opened 2025-11-12
Allow boolean flags for eval jobs

jm424:
Currently flags are passed on as key-value pairs but that approach doesn't extend to boolean flags


RobertTLange:
LGTM! Thank you.

jm424:
Thanks Robert!

PR #56, opened 2025-12-19
feat: Multi-turn agentic architecture

GeorgeWingg:
## Summary

- Adds multi-turn agentic editing and evaluation backends (Codex CLI, Shinka Agent)
- Implements multi-file workspace support with embedding corpus for novelty detection
- Adds bandit sampling integration for agentic mode
- Includes new example configs for boids_flocking_agentic and circle_packing_agentic variants

## Key Changes

- **Agentic backends**: `shinka/edit/agentic.py`, `codex_cli.py`, `shinka_agent.py` - pluggable CLI harnesses that own system prompts and stream events
- **Agentic evaluator**: `shinka/eval/agentic.py` - runs evaluation in agent sessions with metrics extraction
- **Multi-file corpus**: `shinka/core/embedding_corpus.py` - builds embedding text from multiple workspace files
- **Runner integration**: Full async job pipeline with thread-safe parallelism for agentic mode
- **Configs**: New `evolution/agentic.yaml` base config and variant configs

## Test plan

- [x] Unit tests for agentic editor and evaluator (`tests/test_agentic_*.py`)
- [ ] Manual test with circle_packing_agentic config
- [ ] Manual test with boids_flocking_agentic config


zia1138:
@GeorgeWingg
 Curious how well your fork worked. What challenges or advantages did using codex cli have that you noticed? Thanks!

GeorgeWingg:
Hi, the reason I built this was to add the ability to take shinka's evolutionary approach and apply it to real engineering projects, where logic is in multiple files rather than a single file.

Also I have the perspective that it might be valuable for an AI scientist tool to be able to call tools like how the modern agentic models which was not possible with the previous shinka system. 

Cons I would look out for is that the agent can run commands and change files in your file system, and with fork the commands they run are fairly abstracted away, so I would be careful running this on a computer with files you care about.

RobertTLange:
Thank you for the ambitious agentic architecture work. We are pausing this large feature while we clarify supported agents, credentials, workspace containment, and a minimal end-to-end task. Please reopen with a small design-backed vertical slice when that scope is set.

PR #68, opened 2026-02-07
Add Support for Python Virtual Environment Execution

vicruz99:
## Summary
This PR extends ShinkaEvolve's job execution capabilities by adding support for running experiments inside a specific Python virtual environment, similar to the existing Conda environment support.

## Motivation
Currently, ShinkaEvolve supports:
- **Local execution** (using the current Python environment)
- **Conda environments** (via `LocalJobConfig.conda_env` for local jobs and `SlurmCondaJobConfig` for SLURM jobs)
- **Docker containers** (via `SlurmDockerJobConfig`)

However, there's no native support for Python virtual environments (venv/virtualenv), which are commonly used for dependency isolation without requiring Conda. This PR fills that gap by allowing users to specify a Python environment path for job execution.

## Changes

- Added optional `python_path` parameter to `LocalJobConfig` to allow specification of a Python virtual environment path
- Updated documentation on README.md
- Modified __init__.py files to use lazy imports instead of importing all submodules automatically. This change enables shinka to be installed in Python environments (where experiments are executed) without requiring all dependencies to be present. I think shinka really needs to be installed in the python envs where experiments are ran, so the evaluate.py function can be called. Due to this lazy import, shinka can be installed with --no-deps and the env won't be bloated with all the shinka dependencies.

## Usage

- Create a new virtual python environment
- Install the dependencies of your experiment
- Install shinka (without dependencies, to avoid uncessarily bloating your venv):  pip install --no-deps -e .
- In your config file, set your python path like this:

```yaml
job_config:
  python_path: path_to_your_venv/.venv/bin/python

### DIFF
--- a/shinka/launch/scheduler.py
+++ b/shinka/launch/scheduler.py
@@ -141,7 +141,7 @@ class JobScheduler:
                 python_cmd = "python"
                 if self.job_type == "local" and isinstance(self.config, LocalJobConfig):
                     python_cmd = self.config.python_path
-                print("\n\nUsing python command:", python_cmd)
+                #print("\n\nUsing python command:", python_cmd)
                 cmd = [
                     python_cmd,
                     f"{self.config.eval_program_path}",


RobertTLange:
Fixed recently. See Readme file. Supporting venv.

PR #125, opened 2026-04-23
Add Vertex AI for Gemini models

wu375:
## Summary

- add Vertex AI authentication support for Gemini LLM and embedding clients through a shared Google GenAI client helper
- update `shinka_models` provider readiness checks
- add and update related unit tests

## Why

- This PR makes it possible to experiment with Gemini models via Vertex AI auth backend.

## Testing

- Commands run:
  - `uv run shinka_run ...`
- Results:
  - Vertex AI Gemini models working
  - Vertex AI embedding models have caveats (see Risks and compatibility section)

- Commands run:
  - `uv run --with pytest-cov pytest -q -m "not requires_secrets" --cov=shinka --cov-report=term-missing --cov-report=xml:coverage.xml`
- Optional secret-backed tests:
  - `uv run pytest -q -m "requires_secrets"`
- Results:
  - Passed
  - `uv run pytest -q -m "not requires_secrets"`: had 1 warning from `tests/test_bandit_persistence.py` / `shinka/llm/prioritization.py` and appears unrelated to the Vertex AI change

## Risks and compatibility

- Existing Gemini API key behavior is preserved; this change adds an alternate auth path rather than replacing the current one
- Embedding models via Vertex AI
  - This feature does not separate auth for Gemini LLM and embedding models, meaning that when Vertex AI flag is on, both are used via Vertex AI.
  - Vertex AI API takes the raw embedding model name instead of the prefixed `models/<model_name>` form. Based on observed behavior during manual validation, `gemini-embedding-001` did not work on Vertex AI when the prefix was present.
  - For `gemini-embedding-001` on Vertex AI, `embed_content()` appears to work with the raw model name, but `count_tokens()` still does not. This PR preserves the existing fallback behavior when token counting fails, so embeddings can still succeed and cost falls back to the current estimate path.
  - `gemini-embedding-2-preview` appears problematic in Vertex AI testing regardless of the model-name format, so this PR does not attempt to resolve preview-model compatibility beyond the minimal naming change above.
  - Google also introduced the GA `gemini-embedding-2` model today/yesterday. That release may explain some of the preview-model instability, but this PR does not broaden scope beyond Vertex AI compatibility fix.

Thanks for the great work of the Shinka project! Been having fun playing with it :)

### DIFF
--- a/shinka/cli/models.py
+++ b/shinka/cli/models.py
@@ -15,6 +15,7 @@ from shinka.embed.providers.pricing import (
     get_models_by_provider as get_embedding_models_by_provider,
 )
 from shinka.env import load_shinka_dotenv
+from shinka.google_genai import google_genai_auth_mode
 from shinka.llm.providers.pricing import get_all_providers, get_models_by_provider
 
 PROVIDER_ENV_REQUIREMENTS: dict[str, tuple[str, ...]] = {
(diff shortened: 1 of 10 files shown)


RobertTLange:
Thanks @wu375 for the Vertex AI Gemini support and the thorough validation notes. I reviewed the diff and ran the non-secret test suite locally; merging now.

PR #126, opened 2026-04-27
Fix: avoid silent retries when missing Gemini credentials

wu375:
## Summary

- Skip retries for Gemini authentication errors.

## Why

- Gemini calls previously retried every Exception, including credential failures that cannot recover through retry. This made **runs appear stuck** for `shinka_launch`, specifically when GCP (Vertex AI) Application Default Credentials were missing.

Before this change:

- For `shinka_launch`: When Vertex AI is selected with `GOOGLE_GENAI_USE_VERTEXAI=true` but ADC was missing, Gemini client could initialize but fail during the API call.
- The failure entered Gemini’s retry loop and could **spend a long time retrying** the credentials error silently.

After this change:

- Gemini provider calls classify common auth/config failures as non-retryable, skipping exponential retries for those errors.
- Transient errors, such as timeouts, still remain retryable.
- `shinka_launch` shows error quickly (but still stays in the evo loop, which could be a separate problem, not covered by this PR)

## Related Context
Related to the previous PR which added Vertex AI support: https://github.com/SakanaAI/ShinkaEvolve/pull/125

## Testing

- Commands run:
  - `uv run --with pytest-cov pytest -q -m "not requires_secrets" --cov=shinka --cov-report=term-missing --cov-report=xml:coverage.xml`
- Optional secret-backed tests:
  - `uv run pytest -q -m "requires_secrets"`
- Results:
  - All passed

### DIFF
--- a/shinka/llm/providers/gemini.py
+++ b/shinka/llm/providers/gemini.py
@@ -13,6 +13,21 @@ MAX_TRIES = BACKOFF_MAX_TRIES
 MAX_VALUE = BACKOFF_MAX_VALUE
 MAX_TIME = BACKOFF_MAX_TIME
 
+NON_RETRYABLE_GEMINI_ERROR_MARKERS = (
+    "your default credentials were not found",
+    "could not automatically determine credentials",
+    "application default credentials",
+    "google_cloud_project is required",
+    "google_cloud_location is required",
+    "gemini_api_key",
+    "api key not valid",
+    "invalid api key",
+    "api_key_invalid",
+    "unauthenticated",
+    "permission denied",
+    "permission_denied",
+)
+
 
 def build_gemini_thinking_config(thinking_budget: int):
     """Build Gemini ThinkingConfig across SDK versions.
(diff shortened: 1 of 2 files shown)


RobertTLange:
Thanks for tracking this down and putting up the fix. The underlying issue is real: runs should not appear stuck when requested Gemini credentials/config are unavailable.

I opened PR #127 as the follow-up implementation, and I think that should close/supersede this PR. The reason is that the missing-token/config case is better handled before the evo loop starts, using the existing `shinka_models` provider availability workflow, rather than adding a separate Gemini-specific list of auth error strings inside the provider retry path.

PR #127 centralizes the env-based provider availability logic and makes `ShinkaEvolveRunner` fail fast at startup when any requested mutation, meta, novelty, prompt-evolution, or embedding model is unavailable due to missing provider env vars. That covers `GEMINI_API_KEY` mode, Vertex mode (`GOOGLE_GENAI_USE_VERTEXAI + GOOGLE_CLOUD_PROJECT + GOOGLE_CLOUD_LOCATION`), OpenAI/OpenRouter/Azure/Bedrock/etc., and keeps `shinka_models` and runtime startup validation aligned.

So I would close this PR in favor of #127. Runtime retry classification for invalid/expired credentials or permission failures that surface after startup is still useful, but I would handle that separately with structured SDK error fields rather than string matching as part of this PR.

PR #130, opened 2026-05-02
fix: resize bandit state on model changes

### LINKED ISSUE #129: bandit_state.pkl resume crashes with IndexError when llm_models list size changes between runs
> Filed by Claude Code (Sonnet sub-agent of an autonomous research session driven by jeremycochoy). Reproduction context cross-checked by the parent agent before submission.

## Summary

When `shinka_launch` resumes from a saved `bandit_state.pkl` that was created with N entries in `evo_config.llm_models`, but the current invocation specifies M ≠ N models, the bandit's array indexing raises an `IndexError` on the first proposal-generation attempt and the run cannot recover.

## Observed error

```
Error generating proposal for generation <N+1>: index 3 is out of bounds for axis 0 with size 3
```

(Concrete numbers depend on N and M; the message form is `index <≥N> is out of bounds for axis 0 with size <N>`.)

The error fires for every subsequent generation, with no automatic recovery. Stopping and editing `evo_config.llm_models` back to the original N entries unblocks the resume cleanly.

## Where in the source (verified on `main`)

All `set_state` implementations do blind unpickle without checking that the saved arrays match the configured `n_arms`:

- `shinka/llm/prioritization.py`
  - `BanditBase.load_state` (≈ L206–221)
  - `AsymmetricUCB.set_state` (≈ L831–841)
  - `FixedSampler.set_state` (≈ L984–994)
  - `ThompsonSampler.set_state` (≈ L1346–1355)

  Each does `self.<arr> = state["<arr>"].copy()` with no length check or pad/truncate.

- `shinka/core/async_runner.py` — `_load_bandit_state` (≈ L560–577) calls `load_state` and only catches generic `Exception` as a warning. There is no pre-check that `len(state["n_submitted"]) == self.llm_selection.n_arms`.

After load, the per-arm arrays end up at the saved length N. The first arm sampling that yields an index in `[N, M)` hits the IndexError.

`tests/test_bandit_persistence.py` only round-trips with identical `n_arms` — the cross-size case is untested.

## Reproduction

1. Run `shinka_launch` with 3 LLM models for a few generations:

   ```bash
   shinka_launch ... \
     evo_config.llm_models='[openrouter/qwen/qwen3-coder-30b-a3b-instruct,openrouter/qwen/qwen3-30b-a3b,openrouter/qwen/qwen3-30b-a3b-thinking-2507]' \
     evo_config.num_generations=5 \
     output_dir=results/repro_bandit_resize
   ```

2. Wait for at least one generation to complete and confirm `results/repro_bandit_resize/bandit_state.pkl` exists. Stop the run (e.g. `screen -S … -X quit`).

3. Re-run with the **same `output_dir`** but a different number of models (e.g. add two more):

   ```bash
   shinka_launch ... \
     evo_config.llm_models='[openrouter/qwen/qwen3-coder-30b-a3b-instruct,openrouter/qwen/qwen3-30b-a3b,openrouter/qwen/qwen3-30b-a3b-thinking-2507,openrouter/anthropic/claude-sonnet-4.6,openrouter/anthropic/claude-haiku-4.5]' \
     evo_config.num_generations=10 \
     output_dir=results/repro_bandit_resize
   ```

4. Observe `Resuming from generation N` in the log followed immediately by `Error generating proposal for generation N+1: index 3 is out of bounds for axis 0 with size 3` on every subsequent gen.

## Fix idea (one line)

Detect a shape mismatch in `_load_bandit_state` and extend each per-arm array with the class's neutral prior (or truncate, when models are removed) before resuming.

## Environment

- `ShinkaEvolve` `main` branch (verified upstream during the report).
- `shinka_launch` invoked via `--config-dir=./configs task=… evolution@_global_=small_budget_gpu database@_global_=island_small`.
- Hit during a real multi-day run where mid-experiment we wanted to add two new LLMs to the ensemble.

RobertTLange:
## Summary
- Persist `arm_names` with bandit sampler state for name-based resume alignment.
- Resize loaded sampler arrays when `llm_models` expands, shrinks, or reorders.
- Keep legacy `bandit_state.pkl` files without `arm_names` loadable via prefix/index alignment.
- Recompute adaptive observation ranges when the saved arm layout changes so removed arms do not keep influencing resumed selection.
- Preserve cost-aware UCB cost range state across resume and document the fix in `CHANGELOG.md`.

## Why
Resuming a run after changing `llm_models` could leave bandit arrays at the old length, causing posterior/select paths to fail or stale stats to be associated with the wrong model.

Fixes #129.

## Tests
- `.venv/bin/python -m pytest tests/test_bandit_persistence.py tests/test_async_runner_recovery.py -q`
- `.venv/bin/python -m pytest -q`
- `.venv/bin/python -m ruff check shinka/llm/prioritization.py tests/test_bandit_persistence.py`
- `git diff --check HEAD~2..HEAD`
- `.venv/bin/python -m mkdocs build --strict`

## Notes
The targeted pytest run still emits the existing ThompsonSampler decay `RuntimeWarning`, but all tests pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ All notable changes to `shinka-evolve` are documented in this file.
 
 ### Fixed
 
+- Fixed bandit sampler resume from legacy or changed `llm_models` state so saved per-arm arrays are resized, name-aligned when possible, and cost-aware UCB range state remains active after loading `bandit_state.pkl` in PR #130, addressing issue #129.
 - Fixed the WebUI embedding similarity heatmap so hydrated programs with empty embeddings no longer leave the tab stuck on `Loading full embedding data...`, and stale cached full-program data is refetched after summary updates.
 
 ## 0.0.5 - 2026-04-22
--- a/shinka/llm/prioritization.py
+++ b/shinka/llm/prioritization.py
@@ -572,6 +572,32 @@ class AsymmetricUCB(BanditBase):
 
         return cost_ref / cost_denom
 
+    def _restore_cost_range(self, state: Dict[str, Any]) -> None:
+        if (
+            self._state_arm_layout_matches(state, "n_costs")
+            and "min_cost_observed" in state
+            and "max_cost_observed" in state
+        ):
+            self.min_cost_observed = float(state["min_cost_observed"])
+            self.max_cost_observed = float(state["max_cost_observed"])
+            return
+
+        have_cost = self.n_costs > 0.0
+        if not np.any(have_cost):
+            self.min_cost_observed = np.inf
+            self.max_cost_observed = -np.inf
+            return
+
+        mean_costs = self.total_costs[have_cost] / self.n_costs[have_cost]
+        finite_costs = mean_costs[np.isfinite(mean_costs)]
+        if finite_costs.size == 0:
+            self.min_cost_observed = np.inf
+            self.max_cost_observed = -np.inf
+            return
+
+        self.min_cost_observed = float(finite_costs.min())
+        self.max_cost_observed = float(finite_costs.max())
+
     def posterior(self, subset=None, samples=None):
         idx = self._resolve_subset(subset)
         if samples is None or int(samples) <= 1:
@@ -898,6 +924,8 @@ class AsymmetricUCB(BanditBase):
             "obs_min": self._obs_min,
             "n_costs": self.n_costs.copy(),
             "total_costs": self.total_costs.copy(),
+            "min_cost_observed": self.min_cost_observed,
+            "max_cost_observed": self.max_cost_observed,
         }
 
     def set_state(self, state: Dict[str, Any]) -> None:
@@ -916,6 +944,7 @@ class AsymmetricUCB(BanditBase):
         self.total_costs = self._align_state_array(
             state, "total_costs", self.total_costs
         )
+        self._restore_cost_range(state)
 
 
 class FixedSampler(BanditBase):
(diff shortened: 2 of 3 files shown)


RobertTLange:
Follow-up review findings and fixes pushed in `52b42aa` and `568918e`.

Findings addressed:
- Cost-aware UCB resume restored `n_costs` / `total_costs`, but left `min_cost_observed` / `max_cost_observed` at constructor defaults. That made `_normalized_cost_ratio()` return `None` after resume, silently dropping persisted cost-aware selection until a new cost update arrived.
- Legacy `bandit_state.pkl` coverage without `arm_names` only exercised `AsymmetricUCB`, while old states for `ThompsonSampler` and `FixedSampler` use the same compatibility path.

Fixes:
- Persist `min_cost_observed` / `max_cost_observed` in `AsymmetricUCB` state.
- Restore those bounds when the saved/current arm layout still matches.
- Recompute usable cost bounds from retained cost history for legacy or changed-layout state.
- Add regression coverage for cost-aware UCB resume preserving posterior behavior.
- Add legacy no-`arm_names` resize coverage for Thompson and Fixed samplers.

Validation:
- `.venv/bin/python -m pytest tests/test_bandit_persistence.py tests/test_async_runner_recovery.py -q` -> 57 passed
- `.venv/bin/python -m pytest -q` -> 363 passed
- `.venv/bin/python -m ruff check shinka/llm/prioritization.py tests/test_bandit_persistence.py` -> passed
- `git diff --check HEAD~2..HEAD` -> passed
- `.venv/bin/python -m mkdocs build --strict` -> passed

Notes:
- The existing ThompsonSampler decay `RuntimeWarning` is still emitted in the same persistence tests.
- `mypy shinka` is not currently a clean repo gate; it reports existing project-wide typing/stub issues unrelated to this patch.

PR #131, opened 2026-05-03
Add Fortran evolution support

RobertTLange:
## What changed
- Add canonical Fortran language support with f90/f95/f03/f08 aliases, .f90 artifacts, ! EVOLVE markers, and Fortran code fences.
- Teach shinka_run, async validation, failure artifact language inference, and WebUI download/failed-node paths about Fortran.
- Add a distinct Fortran heat-diffusion example with a Python evaluator that compiles candidates with gfortran and scores floating-point stencil checksums.
- Update README, docs, and shinka-setup language tables.

## Why
Fortran is common for numerical/scientific workloads and should be a first-class candidate language rather than an unsupported fallback. The heat-diffusion example demonstrates a compiled numerical stencil workflow rather than duplicating the existing Julia prime-counting task.

## Testing
- uv run pytest tests/test_visualization_meta.py::test_handle_get_program_details_infers_failed_fortran_node_from_suffix tests/test_async_runner_recovery.py::test_get_failure_language_infers_fortran_from_generated_filename tests/test_edit_fortran.py tests/test_async_apply.py::test_validate_code_async_fortran_delegates_to_gfortran -q
- uv run ruff check shinka/core/async_runner.py shinka/webui/visualization.py tests/test_visualization_meta.py tests/test_async_runner_recovery.py tests/test_edit_fortran.py tests/test_async_apply.py
- uv run --with pytest-cov pytest -q -m "not requires_secrets" --cov=shinka --cov-report=term-missing --cov-report=xml:coverage.xml
- uv run mypy --follow-imports=skip --ignore-missing-imports tests/test_*.py tests/conftest.py
- uv run --group docs mkdocs build --strict
- git diff --check

Note: local gfortran is not installed, so the example smoke run was limited to verifying the missing-compiler failure path.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ All notable changes to `shinka-evolve` are documented in this file.
 - Added Vertex AI authentication support for Gemini LLM and embedding clients in PR #125. Thanks @wu375.
 - Added async-runner validation for configured LLM and embedding model environment access before run artifacts are created in PR #127. Thanks @RobertTLange.
 - Added GPT-5.5 and GPT-5.5 Pro entries to the OpenAI LLM pricing catalog.
+- Added Fortran evolution support, including language detection, patch application, validation, visualization metadata, and a compiled heat-diffusion example in PR #131.
 
 ### Fixed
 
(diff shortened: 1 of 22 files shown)


RobertTLange:
Review follow-up pushed in two commits:

1. Finding: `examples/fortran_heat_diffusion/evaluate.py` executed candidate programs without setting `cwd`, so relative file I/O from evolved Fortran code would happen in the caller workspace instead of the evaluator temp dir.
   Solution: `23a297a fix: run fortran evaluation in temp directory` adds `cwd=tmpdir_path` and a regression test covering the subprocess cwd contract.

2. Finding: `tests/test_edit_fortran.py` had intentional SEARCH/REPLACE markers starting at column 1, causing `git diff --check` to report leftover conflict markers.
   Solution: `56a2e31 test: avoid conflict marker false positive` builds the fixture with adjacent string literals so the same patch text is tested without tripping diff hygiene.

Validation run locally:
- `uv run pytest tests/test_fortran_heat_diffusion_example.py tests/test_edit_fortran.py -q` -> 5 passed
- `git diff --check origin/main` -> passed
- Earlier full gate before committing: Ruff, Mypy, and non-secret pytest with coverage passed (`376 passed`).

Note: local machine still has no `gfortran` on PATH, so live compilation can only be validated in CI / an environment with gfortran installed.

PR #134, opened 2026-05-15
feat: add Wolfram Language support

richardassar:
## Summary

Adds Wolfram Language as a first-class evolution target. Four pieces:

- **Language registry** (`shinka/utils/languages.py`): `wolfram` with aliases `wl`/`wls`/`mathematica`, `.wl` extension, `(*` comment prefix, fence tags, EVOLVE-BLOCK marker patterns, block-comment delimiters.
- **Marker validation** (`shinka/edit/marker_validation.py`): `validate_evolve_markers`, hooked into `apply_full_patch` and `apply_diff_patch`. Catches the Wolfram-specific failure where an LLM emits both EVOLVE-BLOCK markers inside one block comment, e.g. `(* EVOLVE-BLOCK-START\nsolve[] := ...\n EVOLVE-BLOCK-END *)` — previously passed the applier (the literal marker strings were present) and silently left `solve[]` undefined. Block-comment languages additionally get strict per-line shape and comment-balance checks; line-comment languages get an existence check; languages without registered marker patterns defer.
- **`wolfram_llm` provider** (`shinka/llm/providers/wolfram_llm.py` + `wolfram_llm_bridge.wl`): model names `wolfram-llm/<Service>/<Model>` for OpenAI, Anthropic, GoogleGemini, DeepSeek, Groq, MistralAI, Cohere, AlephAlpha, TogetherAI. The bridge forwards a role-tagged messages list to `ServiceConnect` + `ServiceExecute["ChatService"]` and extracts real input/output token counts from the response `Usage`, which the provider runs through the project's existing `calculate_cost` against `pricing.csv`. `WOLFRAM_LLM_USE_LLMSYNTHESIZE=1` selects an `LLMSynthesize` + `LLMConfiguration` path instead.
- **Example** (`examples/wolfram_gcd_sum/`): a deoptimised seed for `S(N) = sum_{i,j in 1..N} GCD(i, j)`, structured like `examples/julia_prime_counting/` and `examples/fortran_heat_diffusion/`.

Cross-platform `wolframscript` access (binary resolution, WSL `.exe` fallback, WSL→Win32 interop wrapping) is centralized in `shinka/utils/wolfram.py` and shared by the provider, the `validate_code_async` Wolfram branch, the runner's pre-run availability gate, and the example evaluator.

## Why

Lets ShinkaEvolve target Wolfram Language candidates, and optionally route LLM calls through Wolfram service credentials instead of Python-side API keys.

## Linked issue or context

No tracked issue.

## Testing

Commands run from the repo root:

- `uv run ruff check tests --exclude tests/file.py` — all checks passed
- `uv run mypy --follow-imports=skip --ignore-missing-imports tests/test_*.py tests/conftest.py` — no issues in 57 source files
- `uv run --with pytest-cov pytest -q -m "not requires_secrets" --cov=shinka --cov-report=term-missing --cov-report=xml:coverage.xml` — 509 passed (or 507 + 2 skipped on machines without `wolframscript`)

119 new test cases:

| File | Cases | Covers |
|---|---|---|
| `tests/test_marker_validation.py` | 56 | well-formed and missing markers per language, Wolfram block-comment smuggling repro, line-comment leniency, applier integration |
| `tests/test_wolfram_llm_provider.py` | 51 | model-name parsing, resolver/client wiring, spec construction with multi-turn history, mocked end-to-end round-trip, real-cost computation via `pricing.csv`, error paths, cross-platform argv construction, LLMKit hint, `useLLMSynthesize` routing, the shared helpers |
| `tests/test_edit_wolfram.py` | 4 | Wolfram diff/full patch application and aliases |
| `tests/test_wolfram_bridge.py` | 3 | the bridge run through `wolframscript` against malformed and missing-field input; skipped where the binary is absent |
| `tests/test_wolfram_gcd_sum_example.py` | 3 | the example evaluator's parsing, scoring, and wrong-answer rejection |
| `tests/test_async_apply.py` | +1 | the `validate_code_async` Wolfram branch routes through the shared helpers and escapes candidate paths |
| `tests/test_packaging_release.py` | +1 | the `.wl` bridge is reachable via `importlib.resources` |

The `wolframscript`-gated bridge tests skip on machines without the binary (e.g. `ubuntu-latest` on CI) and run live everywhere else.

## Risks and compatibility

- Additive. No existing provider, language, or applier behaviour changes; the marker-validation hook is a no-op for well-formed candidates.
- Requires `wolframscript` (free Wolfram Engine for Developers, or Mathematica). The provider, the pre-run gate, and the example all fail fast with an actionable message when it is absent. On WSL the helper falls back to `wolframscript.exe` automatically when the Linux name isn't on `PATH`.
- Models not in `pricing.csv` (some Groq/Cohere/AlephAlpha/MistralAI/TogetherAI bare names) report `cost = 0.0` rather than guessing.
- `subprocess.TimeoutExpired` / `asyncio.TimeoutError` retries are capped at 2 (vs. the 20 used for other retryable errors), bounding a hung bridge call.
- The `LLMSynthesize` path (`WOLFRAM_LLM_USE_LLMSYNTHESIZE=1`) is not exercised by automated tests — the default `ServiceExecute["ChatService"]` path is. Because `LLMSynthesize` takes a single input string, that path collapses history into a labelled transcript; the default path forwards real multi-turn history.

- **Example:** `examples/wolfram_gcd_sum` — speed up a deoptimised Wolfram solver for `S(300) = sum_{i,j in 1..300} GCD(i, j)` (correct answer 336784).
- **Baseline:** the deoptimised seed, timed by the evaluator on every run through the same `RepeatedTiming` harness as the candidate. `initial.wl` is located via `Path(__file__).parent`, so calibration works regardless of where the candidate program lives during a run.
- **Metric:** `combined_score = baseline_time / median(candidate_time)`; candidates returning the wrong integer or timing out score `-1.0`.
- **Command:** `cd examples/wolfram_gcd_sum && python run_evo.py`, with `GEMINI_API_KEY` set (default model `wolfram-llm/GoogleGemini/gemini-2.5-flash-lite`).
- **Result:** verified end-to-end against Gemini on WSL Ubuntu with Wolfram Engine 14.3. Candidates are applied through the Wolfram patch path, executed through `wolframscript`, scored as speedup over the calibrated baseline, and accepted or rejected. Wrong-answer / timeout candidates are scored `-1.0`; a transient Gemini rate-limit was retried and recovered by the existing `@backoff` handler.

## Candidate future examples

Possible follow-up Wolfram examples, each with a clear scalar score, ideally a continuous gradient, plus a correctness gate.

### Symbolic quality (non-runtime scoring)

- Simplify an expression — minimize `LeafCount` and/or expression depth while `FullSimplify`-equivalence to the original is the correctness gate.
- Find a closed form — given a sum/integral/recurrence, evolve toward an expression with no `Sum`/`Integrate`/`HoldForm` heads left; score on residual symbolic complexity.
- Rewrite-rule discovery — evolve a `ReplaceAll` rule set that normalizes a family of expressions; score on how many test expressions reach canonical form.

### Numerical accuracy scoring

- Function approximation — evolve an elementary-function approximant minimizing max error over an interval (classic Remez-style target, but open-ended).
- Quadrature / ODE — evolve an integration scheme; score on error vs. a high-precision reference at fixed work budget.
- Curve / model fitting — evolve a model form for a dataset; score on cross-validated residual.

### Runtime (same family as the current example, different shape)

- Memory/allocation rather than wall-clock — score on `MaxMemoryUsed`.
- Asymptotic-scaling task — score across several `N` so only genuine complexity wins show up, not constant-factor luck.

### Combinatorial / search

- Geometry packing or tiling — Wolfram's geometry primitives make a clean evaluator; score on packing density or count.
- Constraint-satisfaction encoding — evolve a `FindInstance`/`Solve` formulation; score on solve time or solutions found.

### Multi-objective

- Pareto-style: accuracy *and* runtime, or correctness *and* `LeafCount` — shows ShinkaEvolve handling a non-scalar trade-off.

## Docs and UI

- `CHANGELOG.md` updated under the `## TBD` section.
- `examples/wolfram_gcd_sum/README.md` added; the example is registered in `README.md`, `docs/examples.md`, and `docs/getting_started.md`.
- No WebUI or CLI behaviour changed.


RobertTLange:
Thanks for the detailed implementation and tests here. I’m not comfortable merging this in its current shape because it combines several separable changes, and the largest one adds a lot of maintenance surface that does not seem necessary for Wolfram Language evolution support.

The part that may make sense for Shinka core is minimal Wolfram-as-target-language support: language registry entries, `.wl` fences/extensions, marker handling, and perhaps one small optional example. But the `wolfram_llm` provider feels like a different feature. It is not Wolfram-language-specific; it is a new LLM routing layer that shells out through `wolframscript`, maps Shinka model names into `wolfram-llm/<Service>/<Model>`, translates credentials for many providers, packages and maintains a `.wl` bridge, depends on Wolfram-side service paclets/auth behavior, and then re-enters the same OpenAI/Anthropic/Gemini/etc. ecosystem that Shinka already supports directly.

That seems too broad and vendor-specific for core. A Wolfram candidate can be generated by the existing providers; in fact one of the new examples already uses `claude-cli/sonnet`, and the GCD example could use Shinka’s existing Gemini path rather than `wolfram-llm/GoogleGemini/...`. If users specifically want to route LLM calls through Wolfram credentials, I think that is better as an external plugin, recipe, or follow-up integration with a stronger general justification.

I’m also concerned about the marker-validation change as written. The goal is reasonable, but the implementation globally balances block comment delimiters for both Wolfram and Markdown using a raw character scan. That can reject otherwise valid candidate text, for example Wolfram strings containing `(*`, or Markdown/HTML comment delimiters unrelated to the evolve markers. Since this hooks into the core patch appliers, false positives would affect existing non-Wolfram workflows too.

My suggested path:

1. Split this PR.
2. Drop `wolfram_llm` from core, or move it outside the main package.
3. Keep only a narrow Wolfram language-support PR if we want this target: registry/fence/extension support, targeted marker handling, parse validation, and tests.
4. Rework marker validation so it validates the evolve marker structure itself without globally parsing arbitrary block comments in unrelated content.
5. Consider keeping just one Wolfram example, or hosting the examples externally if they are mainly Wolfram-specific demos.

The implementation effort is clear, but I think the current scope moves Shinka toward problem/vendor-specific integrations rather than a lean evolution framework.

richardassar:
The PR is now narrowed to Wolfram language target support. The two original commits you reviewed (`bab969a`, `5f51a70`) are unchanged at the base, and the response is three commits on top, one per point.

Actioned, per your suggested path:

1. **Marker validation reworked** (`c89bf94`). The global block comment delimiter balancing is gone. Validation is now scoped to the marker lines themselves: for block comment languages each marker must match its strict canonical form (e.g. `(* EVOLVE-BLOCK-START *)`), which still catches a marker emitted without its closing delimiter, but it never parses or balances comments anywhere else in the file. A `(*` inside an unrelated Wolfram string literal, or HTML comment delimiters, can no longer cause a false rejection in the core appliers. A test covers exactly that case.

2. **`wolfram_llm` provider removed from this PR** (`420a915`). It is an LLM routing layer rather than language target support, so it is out of this PR entirely: provider, `.wl` bridge, model name wiring, packaging, and tests. It is preserved on a separate branch on our fork (`wolfram-llm-provider`) in case it is useful later as an external plugin or integration, but it is not proposed for core here.

3. **One example, existing provider path** (`a99a5d9`). The PR keeps a single example, `wolfram_gcd_sum`, and it now uses Shinka's existing `gemini-2.5-flash-lite` path directly rather than routing through the provider. The second example (`wolfram_simplify`) is moved to a separate branch on our fork (`wolfram-simplify-example`).

The PR is now narrowly: language registry entries, `.wl` fences and extension, marker handling, `wolframscript` parse validation, one example, and tests. A language backend, with nothing vendor specific. Checks pass (ruff, mypy, `pytest -m "not requires_secrets"`, 451 passed).

Where these pieces now live on our fork:

- `wolfram-language-support`: this PR. Wolfram as an evolution target (language registry, `.wl` fences and extension, marker handling and validation, `wolframscript` parse validation, one example, tests).
- `wolfram-llm-provider`: the `wolfram_llm` provider that was removed from this PR. Routes LLM calls through `wolframscript`'s service framework. Kept available as a possible external plugin or integration.
- `wolfram-simplify-example`: the second example removed from this PR, `wolfram_simplify`. Expression simplification scored by `LeafCount`.
- `claude-cli-oauth-provider`: a separate `claude_cli` provider (LLM calls via the Claude Code OAuth login). Independent of the Wolfram work.

RobertTLange:
Thanks for narrowing this PR after the previous review. I did a follow-up pass and found two concrete issues, then prepared fixes in a separate branch because the PR head branch has `maintainerCanModify=false` and GitHub rejected a direct push to `WolframInstitute:wolfram-language-support` with 403.

Fixes pushed to `SakanaAI:pr-134-review-fixes`:

- `6ebc5df fix: preserve Wolfram markers for full patches`
  - Fixes markerless Wolfram full patches. The editable range detection now treats Wolfram/Markdown block-comment marker delimiters as part of the marker boundary, so a full patch that omits EVOLVE markers no longer reassembles malformed `(* EVOLVE-BLOCK-START` / `EVOLVE-BLOCK-END *)` lines.
  - Also keeps the malformed single-comment marker case rejected before range extraction.
  - Validation: `uv run pytest -q tests/test_edit_wolfram.py::test_apply_full_patch_supports_wolfram_without_markers tests/test_marker_validation.py::test_apply_full_patch_rejects_wolfram_smuggled_markers tests/test_edit_wolfram.py tests/test_marker_validation.py tests/test_edit_base.py tests/test_edit_fortran.py tests/test_edit_julia.py` passed.

- `15a2f31 fix: handle malformed Wolfram evaluator output`
  - Fixes the evaluator path where `wolframscript` exits 0 but leaves empty, invalid, or incomplete JSON. Those cases now produce normal failed `metrics.json` / `correct.json` output instead of crashing the evaluator with `JSONDecodeError` or `KeyError`.
  - Validation: `uv run pytest -q tests/test_wolfram_gcd_sum_example.py` passed.

Combined validation after both fixes:

- `uv run pytest -q tests/test_edit_wolfram.py tests/test_marker_validation.py tests/test_async_apply.py tests/test_wolfram_gcd_sum_example.py` -> 71 passed.
- `uv run ruff check shinka/edit/apply_diff.py shinka/edit/apply_full.py examples/wolfram_gcd_sum/evaluate.py tests/test_edit_wolfram.py tests/test_wolfram_gcd_sum_example.py` -> passed.

Could you please confirm whether you have run `examples/wolfram_gcd_sum` end-to-end with `python run_evo.py` and verified that the example works with a real `wolframscript` installation and the configured Gemini credentials?

RobertTLange:
Follow-up on the fixes above: the branch to merge or cherry-pick from is `SakanaAI:pr-134-review-fixes`.

It contains:

- `6ebc5df fix: preserve Wolfram markers for full patches`
- `15a2f31 fix: handle malformed Wolfram evaluator output`

These commits are not visible in this PR because the PR head is `WolframInstitute:wolfram-language-support` and maintainer edits are disabled, so GitHub rejected a direct push to the contributor branch. If you merge/cherry-pick `SakanaAI:pr-134-review-fixes` into `WolframInstitute:wolfram-language-support`, the fixes will appear in PR #134.

richardassar:
Both of your fixes are in the PR branch. I advanced `WolframInstitute:wolfram-language-support` to `SakanaAI:pr-134-review-fixes`, so `6ebc5df` and `15a2f31` now appear in PR #134 with your authorship intact.

On your question: yes, `examples/wolfram_gcd_sum` has been run end to end with `python run_evo.py` against a real `wolframscript` installation (Wolfram Engine 14.3) and the configured Gemini credentials, on the native `gemini-2.5-flash-lite` path.

What the run showed:

1. The loop runs cleanly. Gemini generates a mutation, it is applied through the Wolfram patch path, executed through `wolframscript`, scored, and accepted or rejected. Candidates that returned a wrong integer were scored `-1.0` and rejected.

2. The run surfaced one issue in the gcd_sum harness, now fixed in commit `d245449`. The harness times `solve[]` with `RepeatedTiming`, which averages repeated evaluations. The first evaluation in a fresh kernel pays a startup cost (lazy loading of the builtins it touches) that later evaluations do not. That cold run could exceed `RepeatedTiming`'s internal time threshold, so it stopped after a single slow run and reported it as the average, inflating every candidate's measured speedup (the unchanged seed scored about 16x against itself). The fix runs `solve[]` once, discarded, before the timed `RepeatedTiming`, so it measures steady state. After it the seed scores about 1.0 against itself (baseline 450 ms, candidate median 463 ms).

The PR branch now has your two fixes plus `d245449`. Checks pass: ruff, mypy, and `pytest -m "not requires_secrets"`.

RobertTLange:
Thanks for the work on this and for the quick communication!

richardassar:
You're welcome, thank you! 😊

Issue #143, opened 2026-07-05
Unbounded LLM-call stall wedges a run for hours-to-days; stuck-system watchdog cannot see it (repro attached)


MateoTTR:
## Summary

A single stalled (or very slow) OpenAI-compatible endpoint wedges an entire evolution run mid-generation, with zero forward progress, for a worst case of **~60 hours per proposal** — and the built-in stuck-system watchdog structurally cannot detect it. Reproduced model-independently on shinka 0.0.7 with a fake endpoint that accepts the HTTP request and never sends a response byte.

## Root cause — a four-layer multiplicative retry stack with no external bound

One proposal is one `await self.llm.query(...)` with no `asyncio.wait_for` at any call site (`shinka/core/async_runner.py:3606` in `_run_patch_async`; `:3382` in `_run_fix_patch_async`; `:1817` in `_generate_initial_program`). The only bound on the await is the HTTP client's own timeout — and the layers around it multiply:

| Layer | Where | Bound | Worst case vs. a fully stalled endpoint |
|---|---|---|---|
| L1 — client timeout | `shinka/llm/client.py:182`, `shinka/llm/constants.py:1` (`TIMEOUT = 1200`, hard-coded, no config/env override) | 1200 s per HTTP attempt | — |
| L1b — openai SDK internal retries | SDK default `max_retries=2`, never overridden | 3 attempts per `create()` | ≈ 60 min per `create()` |
| L2 — `backoff` decorator | `shinka/llm/providers/local_openai.py:105-117` (`max_tries=20`, `max_time=6000`) | elapsed ≥ 6000 s | ≈ 2 h per query call |
| L3 — `AsyncLLMClient.query` retry loop | `shinka/llm/llm.py:602-621` (`MAX_RETRIES = 3`) | 3 tries | ≈ 6 h, then returns `None` |
| L4 — patch-attempt loop | `shinka/core/async_runner.py:3604` (`max_patch_attempts`, commonly raised to ~10 for weaker models) | per config | **≈ 60 h per proposal** |

Two aggravations:

- **The trickle variant is infinite**: httpx's read timeout is per-received-chunk, so an endpoint that dribbles bytes resets L1 forever and no outer layer ever fires.
[...]

PR #144, opened 2026-07-07
Bound OpenAI-compatible LLM retry stalls with configurable budgets

dexhunter:
## Summary

- Expose environment overrides for LLM request timeout, Shinka retry count, OpenAI SDK retry count, and provider backoff budgets.
- Route OpenAI-compatible clients through `OPENAI_MAX_RETRIES`, defaulting to `0`, so SDK retries do not silently stack under Shinka's own retry/backoff layers.
- Add a local stalling OpenAI-compatible endpoint regression test that verifies configured retry bounds produce one request and a fast `APITimeoutError`.

## Motivation

This addresses the configurable timeout/retry-budget portion of #143. Today a stalled OpenAI-compatible endpoint can hold an evolution run because `TIMEOUT`, provider `backoff` budgets, and Shinka's outer LLM retry count are hard-coded, while OpenAI-compatible clients also keep SDK retries enabled by default.

This PR keeps the long default request timeout intact for normal science-generation workloads, but allows users and tests to bound stalled local/OpenAI-compatible providers explicitly.

## Validation

- `uvx ruff check --fix shinka/llm/constants.py shinka/llm/client.py shinka/llm/llm.py tests/test_llm_retry_constants.py tests/test_llm_client_backends.py tests/test_llm_stall_timeout.py`
- `uv run pytest tests/test_llm_client_backends.py tests/test_llm_retry_constants.py tests/test_llm_stall_timeout.py tests/test_headless_provider.py -q`
  - `24 passed`
- External wall-clock eval against a fake endpoint that accepts a request and never responds:
  - clean `main`: `60.0s` capped by the eval timeout
  - this patch: `2.210s`, `request_count=1`, no retry penalty

Note: I also tried full-repo `uvx ruff check --fix .`; current `main` has unrelated pre-existing lint failures in examples/plotting/test fixture files, so I kept the final lint run scoped to the files changed here.

## Autoresearch

Supplementary autoresearch: [20-step public dashboard](https://dashboard.weco.ai/share/VkBbyn4cd_AsWwa6ObmENCloa97rKc4P).

### DIFF
--- a/shinka/llm/constants.py
+++ b/shinka/llm/constants.py
@@ -5,16 +5,19 @@ from shinka.env import load_shinka_dotenv
 load_shinka_dotenv()
 
 
-def _env_int(name: str, default: int) -> int:
+def _env_int(name: str, default: int, *, minimum: int = 1) -> int:
     value = os.getenv(name)
     if value is None or value == "":
         return default
-    return int(value)
+    parsed = int(value)
+    if parsed < minimum:
+        raise ValueError(f"{name} must be >= {minimum}, got {parsed}")
+    return parsed
 
 
 TIMEOUT = _env_int("SHINKA_LLM_TIMEOUT", 1200)
 MAX_RETRIES = _env_int("SHINKA_LLM_MAX_RETRIES", 3)
-OPENAI_MAX_RETRIES = _env_int("SHINKA_OPENAI_MAX_RETRIES", 0)
+OPENAI_MAX_RETRIES = _env_int("SHINKA_OPENAI_MAX_RETRIES", 0, minimum=0)
 BACKOFF_MAX_TRIES = _env_int("SHINKA_LLM_BACKOFF_MAX_TRIES", 20)
 BACKOFF_MAX_VALUE = _env_int("SHINKA_LLM_BACKOFF_MAX_VALUE", 20)
 BACKOFF_MAX_TIME_MULTIPLIER = _env_int("SHINKA_LLM_BACKOFF_MAX_TIME_MULTIPLIER", 5)
--- a/shinka/llm/constants.py
+++ b/shinka/llm/constants.py
@@ -1,5 +1,24 @@
-TIMEOUT = 1200
-BACKOFF_MAX_TRIES = 20
-BACKOFF_MAX_VALUE = 20
-BACKOFF_MAX_TIME_MULTIPLIER = 5
-BACKOFF_MAX_TIME = TIMEOUT * BACKOFF_MAX_TIME_MULTIPLIER
+import os
+
+from shinka.env import load_shinka_dotenv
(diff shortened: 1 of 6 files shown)


shinka/llm/constants.py
```diff
@@ -1,5 +1,24 @@
...
+
+def _env_int(name: str, default: int) -> int:
+    value = os.getenv(name)
+    if value is None or value == "":
+        return default
+    return int(value)
+
+
+TIMEOUT = _env_int("SHINKA_LLM_TIMEOUT", 1200)
+MAX_RETRIES = _env_int("SHINKA_LLM_MAX_RETRIES", 3)
```
RobertTLange, line 19: `SHINKA_LLM_MAX_RETRIES` is used as the total attempt count in `LLMClient` / `AsyncLLMClient` loops, not as SDK-style retries. If someone sets this to `0` to disable the outer retry layer, the loop never calls the provider and returns `None` silently. Please either validate this as `>= 1` or derive total attempts from a retry count, and add a high-level regression for the `0` case.

shinka/llm/constants.py
```diff
@@ -1,5 +1,24 @@
...
+OPENAI_MAX_RETRIES = _env_int("SHINKA_OPENAI_MAX_RETRIES", 0)
+BACKOFF_MAX_TRIES = _env_int("SHINKA_LLM_BACKOFF_MAX_TRIES", 20)
```
RobertTLange, line 21: These newly configurable retry/time budgets should reject invalid bounds. For example, `BACKOFF_MAX_TRIES=0` does not cleanly mean “no retries” with `backoff`; it can still retry until `max_time`, and zero/negative timeout or backoff values can produce surprising SDK/backoff behavior. Please validate timeout/backoff/top-level attempts as positive values, while allowing `SHINKA_OPENAI_MAX_RETRIES=0` as the intentional “single Shinka retry layer” setting.

shinka/llm/constants.py
dexhunter: Fixed in 7790125. SHINKA_LLM_MAX_RETRIES now fails fast below 1, and the regression covers zero so the outer loop cannot silently skip all provider calls.

shinka/llm/constants.py
dexhunter: Fixed in 7790125. Timeout, backoff tries/value/time/multiplier, and top-level attempts now require positive values; SHINKA_OPENAI_MAX_RETRIES intentionally allows 0 but rejects negatives. Added subprocess import-time regressions for every bound. Focused result: 22 tests passed; Ruff check/format and diff check passed.

RobertTLange:
Thank you so much! LGTM. Will be part of the next release.

PR #145, opened 2026-07-11
fix(launcher): clean up local evaluator descendants

### LINKED ISSUE #143: Unbounded LLM-call stall wedges a run for hours-to-days; stuck-system watchdog cannot see it (repro attached)

dexhunter:
## Summary

- Keep each POSIX evaluator process group owned by a live supervisor until cleanup, even when the evaluator leader exits first.
- Report the evaluator exit code through a private pipe while preserving the existing ProcessWithLogging interface.
- Retain group ownership after SIGTERM so cancellation can escalate safely to SIGKILL without signalling a reused process-group ID.
- Track executor-backed submissions through caller cancellation; late jobs stay runner-owned until cancellation or result reaping is confirmed, and their evaluation slot remains reserved until then.
- Add raw-job-ID status checks plus process-tree, repeated-cancellation, late-result, timeout, and shutdown regressions.

## Why

Local evaluators can launch workers of their own. Killing only the direct child leaves those workers alive. A second race occurs when a proposal task is cancelled while its executor-backed evaluation submission is still in progress: the submission can complete later without being tracked, leaving both a live job and a permanently occupied evaluation slot.

The process group must remain owned after its leader exits, and late-submission cleanup must outlive repeated cancellation of the proposal task. This patch makes both lifetimes explicit and releases ownership only after the corresponding process or scheduler job has been reaped.

## Linked issue or context

- Fixes #143. PR #144 handled the retry-budget portion separately.
- The initial process-tree direction came from [autoresearch with Weco](https://dashboard.weco.ai/share/qD-HzpDHi_61B5cMNFZF_ltXRyUUAQIe); the review-race fixes were rebuilt and validated manually.

## Testing

Required base: current main at 5bb134a.

- Current-base paired process-tree probe: leader_exit_orphan_descendants=1 on main, 0 on this branch.
- Current-base paired cancellation probe: late_jobs_created=1 late_jobs_cancelled=0 slots_in_use=1 on main; late_jobs_created=1 late_jobs_cancelled=1 slots_in_use=0 on this branch.
- uv run --python 3.11 pytest -q tests/test_local_process_cleanup.py tests/test_async_runner_recovery.py: 50 passed.
- uv run pytest -q: 717 passed, with 5 existing bandit-persistence warnings.
- uv run ruff check tests --exclude tests/file.py: passed.
- uv run ruff check shinka/core/async_runner.py shinka/launch/local.py shinka/launch/scheduler.py: passed.
- uv run mypy --follow-imports=skip --ignore-missing-imports tests/test_*.py tests/conftest.py: no issues in 66 source files.
- Changed-source mypy reports the same 35 existing errors on this branch and current main; no new error was introduced.
- uv run black --check --target-version py311 shinka/launch/local.py tests/test_local_process_cleanup.py: passed.
- git diff --check and diff secret/forbidden-term scans: passed.
- uvx ruff check --fix . was run as required; it reports 11 unrelated pre-existing repository violations after four unrelated auto-fixes, which were reverted. The scoped checks above pass.

## Risks and compatibility

- Windows behavior is unchanged and still targets only the direct child.
- POSIX descendants that deliberately detach into a separate session remain outside the evaluator process group.
- The supervisor is internal. Manually constructed ProcessWithLogging wrappers continue to use direct-child behavior unless group ownership is explicitly provided.
- A failed scheduler cancellation keeps the evaluation slot reserved and retries. If the job has already finished, the result path is reaped before the slot is released.

- Primary metric: leader_exit_orphan_descendants (minimize), from 1 on current main to 0.
- Cancellation invariant: late submitted jobs are cancelled or reaped before their evaluation slot is released.
- Ordinary completed jobs, timeout cleanup, scheduler cancellation, graceful termination, escalation, repeated cancellation, and runner shutdown are covered.

## Docs and UI

- No user-facing CLI or WebUI behavior changed.

### DIFF
--- a/shinka/core/async_runner.py
+++ b/shinka/core/async_runner.py
@@ -164,6 +164,7 @@ class AsyncRunningJob:
     completion_detected_at: Optional[float] = None
     discard_if_completed: bool = False
     evaluation_slot_released: bool = False
+    cleanup_cancel_attempted: bool = False
 
 
 @dataclass
@@ -5471,6 +5472,44 @@ class ShinkaEvolveRunner:
 
         logger.info("Meta summarizer task exited")
 
+    async def _cancel_running_jobs_for_cleanup(self) -> None:
+        """Cancel each evaluator still owned by the runner exactly once."""
+        surviving_jobs: List[AsyncRunningJob] = []
+        cancelled_count = 0
+
+        for job in list(self.running_jobs):
+            if job.cleanup_cancel_attempted:
+                surviving_jobs.append(job)
+                continue
+
+            job.cleanup_cancel_attempted = True
+            try:
+                cancelled = await self.scheduler.cancel_job_async(job.job_id)
+            except Exception as e:
+                logger.warning(
+                    f"Error cancelling active job {job.job_id} "
+                    f"(gen {job.generation}) during cleanup: {e}"
+                )
+                surviving_jobs.append(job)
+                continue
+
+            if cancelled:
+                self.submitted_jobs.pop(str(job.job_id), None)
+                await self._release_evaluation_slot_once(job)
+                cancelled_count += 1
+            else:
+                logger.warning(
+                    f"Failed to cancel active job {job.job_id} "
+                    f"(gen {job.generation}) during cleanup; keeping it tracked"
+                )
+                surviving_jobs.append(job)
+
+        self.running_jobs = surviving_jobs
+        if cancelled_count:
+            logger.info(
+                f"Cancelled {cancelled_count} active evaluation job(s) during cleanup"
+            )
+
     async def _cleanup_async(self):
         """Cleanup async resources."""
         try:
@@ -5484,6 +5523,8 @@ class ShinkaEvolveRunner:
                 await asyncio.gather(
                     *self.active_proposal_tasks.values(), return_exceptions=True
                 )
+
+            await self._cancel_running_jobs_for_cleanup()
 
             # Final recomputation of prompt percentiles to ensure fitness is accurate
             if self.prompt_db is not None and self.db is not None:
--- a/shinka/core/async_runner.py
+++ b/shinka/core/async_runner.py
@@ -38,6 +38,7 @@ from shinka.llm import (
 )
 from shinka.embed import AsyncEmbeddingClient
 from shinka.launch import JobScheduler, JobConfig, LocalJobConfig
+from shinka.launch.scheduler import SchedulerSubmissionCleanupError
 from shinka.edit.async_apply import (
     apply_patch_async,
     get_code_embedding_async,
@@ -5152,6 +5153,24 @@ class ShinkaEvolveRunner:
             job_id = await asyncio.shield(submission_task)
         except asyncio.CancelledError:
             raise
+        except SchedulerSubmissionCleanupError as e:
+            getattr(self, "_pending_evaluation_submissions", {}).pop(
+                submission_task, None
+            )
+            if e.cleanup_succeeded:
+                await self.evaluation_slot_pool.release(evaluation_worker_id)
+                return
+            job_id = e.job_id
+            late_jobs = getattr(self, "_late_submission_jobs", None)
+            if late_jobs is None:
+                late_jobs = {}
+                self._late_submission_jobs = late_jobs
+            late_jobs[id(job_id)] = job_id
+            logger.error(
+                f"Scheduler shutdown cleanup did not finish for evaluation job "
+                f"{job_id}; retaining its evaluation slot"
+            )
+            return
         except Exception as e:
             logger.warning(f"Evaluation submission failed after cancellation: {e}")
             getattr(self, "_pending_evaluation_submissions", {}).pop(
@@ -5698,6 +5717,9 @@ class ShinkaEvolveRunner:
         """Cleanup async resources."""
         try:
             deadline = time.monotonic() + self._get_cleanup_retry_timeout()
+            begin_scheduler_shutdown = getattr(self.scheduler, "begin_shutdown", None)
+            if begin_scheduler_shutdown is not None:
+                begin_scheduler_shutdown()
 
             # Cancel remaining proposal tasks
             for task in self.active_proposal_tasks.values():
@@ -5741,9 +5763,6 @@ class ShinkaEvolveRunner:
                     timeout=max(0.0, deadline - time.monotonic()),
                 )
                 if pending:
-                    for task in pending:
-                        task.cancel()
-                    await asyncio.sleep(0)
                     late_job_ids = [
                         str(job_id)
                         for job_id in getattr(
--- a/shinka/core/async_runner.py
+++ b/shinka/core/async_runner.py
@@ -77,6 +77,10 @@ from shinka.utils.languages import get_evolve_comment_prefix
 
 logger = logging.getLogger(__name__)
 
+_CLEANUP_RETRY_TIMEOUT_SECONDS = 10.0
+_CLEANUP_RETRY_INITIAL_DELAY_SECONDS = 0.05
+_CLEANUP_RETRY_MAX_DELAY_SECONDS = 1.0
+
 
 def _print_gradient_logo_and_mirror(
     log_path: Optional[Path] = None,
@@ -164,7 +168,6 @@ class AsyncRunningJob:
     completion_detected_at: Optional[float] = None
     discard_if_completed: bool = False
     evaluation_slot_released: bool = False
-    cleanup_cancel_attempted: bool = False
 
 
 @dataclass
@@ -536,6 +539,7 @@ class ShinkaEvolveRunner:
         self.proposal_queue = asyncio.Queue()
         self.active_proposal_tasks: Dict[str, asyncio.Task] = {}
         self._submission_cleanup_tasks: Set[asyncio.Task] = set()
+        self._late_submission_jobs: Dict[int, Union[str, Any]] = {}
 
         # Performance tracking
         self.total_proposals_generated = 0
@@ -5062,6 +5066,35 @@ class ShinkaEvolveRunner:
         tasks.add(task)
         task.add_done_callback(tasks.discard)
 
+    def _get_cleanup_retry_timeout(self) -> float:
+        return getattr(
+            self, "_cleanup_retry_timeout_seconds", _CLEANUP_RETRY_TIMEOUT_SECONDS
+        )
+
+    async def _try_cleanup_job(self, job_id: Any, results_dir: str) -> bool:
+        """Cancel a job, or confirm that it is terminal and reap it."""
+        try:
+            if await self.scheduler.cancel_job_async(job_id):
+                return True
+        except Exception as e:
+            logger.warning(f"Error cancelling evaluation job {job_id}: {e}")
+
+        try:
+            still_running = await self.scheduler.check_job_id_status_async(job_id)
+        except Exception as e:
+            logger.warning(f"Could not confirm state of evaluation job {job_id}: {e}")
+            return False
+
+        if still_running:
+            return False
+
+        try:
+            await self.scheduler.get_job_results_async(job_id, results_dir)
+        except Exception as e:
+            logger.warning(f"Could not reap terminal evaluation job {job_id}: {e}")
+            return False
+        return True
+
     async def _cleanup_late_evaluation_submission(
         self,
         submission_task: asyncio.Task,
@@ -5076,52 +5109,45 @@ class ShinkaEvolveRunner:
             await self.evaluation_slot_pool.release(evaluation_worker_id)
             return
 
-        retry_delay = 0.05
-        while True:
-            try:
-                cancelled = await self.scheduler.cancel_job_async(job_id)
-            except Exception as e:
-                logger.warning(
-                    "Error cancelling evaluation submitted after proposal "
-                    f"cancellation: {e}"
-                )
-            else:
-                if cancelled:
-                    await self.evaluation_slot_pool.release(evaluation_worker_id)
-                    return
+        late_jobs = getattr(self, "_late_submission_jobs", None)
+        if late_jobs is None:
+            late_jobs = {}
+            self._late_submission_jobs = late_jobs
+        late_job_key = id(job_id)
+        late_jobs[late_job_key] = job_id
 
-                try:
-                    still_running = await self.scheduler.check_job_id_status_async(
-                        job_id
-                    )
-                except Exception as e:
-                    logger.warning(
-                        "Could not confirm the state of an evaluation submitted "
-                        f"after proposal cancellation: {e}"
-                    )
-                else:
-                    if not still_running:
-                        try:
-                            await self.scheduler.get_job_results_async(
-                                job_id, results_dir
-                            )
-                        except Exception as e:
-                            logger.warning(
-                                "Could not reap an evaluation submitted after "
-                                f"proposal cancellation: {e}"
-                            )
-                        else:
-                            await self.evaluation_slot_pool.release(
-                                evaluation_worker_id
-                            )
-                            return
+        deadline = time.monotonic() + self._get_cleanup_retry_timeout()
+        retry_delay = _CLEANUP_RETRY_INITIAL_DELAY_SECONDS
+        while True:
+            remaining = deadline - time.monotonic()
+            if remaining <= 0:
+                break
+            try:
+                cleaned = await asyncio.wait_for(
+                    self._try_cleanup_job(job_id, results_dir), timeout=remaining
+                )
+            except asyncio.TimeoutError:
+                break
+
+            if cleaned:
+                late_jobs.pop(late_job_key, None)
+                await self.evaluation_slot_pool.release(evaluation_worker_id)
+                return
 
             logger.warning(
                 "Evaluation submitted after proposal cancellation remains owned; "
                 "retrying cleanup"
             )
-            await asyncio.sleep(retry_delay)
-            retry_delay = min(1.0, retry_delay * 2)
+            remaining = deadline - time.monotonic()
+            if remaining <= 0:
+                break
+            await asyncio.sleep(min(retry_delay, remaining))
+            retry_delay = min(_CLEANUP_RETRY_MAX_DELAY_SECONDS, retry_delay * 2)
+
+        logger.error(
+            f"Could not confirm cleanup of late evaluation job {job_id} before "
+            "the cleanup deadline; retaining its evaluation slot"
+        )
 
     async def _submit_evaluation_job_with_slot(
         self,
@@ -5556,41 +5582,52 @@ class ShinkaEvolveRunner:
         logger.info("Meta summarizer task exited")
 
     async def _cancel_running_jobs_for_cleanup(self) -> None:
-        """Cancel each evaluator still owned by the runner exactly once."""
-        surviving_jobs: List[AsyncRunningJob] = []
-        cancelled_count = 0
+        """Retry cleanup of active evaluators until success or a fixed deadline."""
+        pending_jobs = list(self.running_jobs)
+        cleaned_count = 0
+        deadline = time.monotonic() + self._get_cleanup_retry_timeout()
+        retry_delay = _CLEANUP_RETRY_INITIAL_DELAY_SECONDS
 
-        for job in list(self.running_jobs):
-            if job.cleanup_cancel_attempted:
-                surviving_jobs.append(job)
-                continue
+        while pending_jobs:
+            surviving_jobs: List[AsyncRunningJob] = []
+            for job in pending_jobs:
+                remaining = deadline - time.monotonic()
+                if remaining <= 0:
+                    surviving_jobs.append(job)
+                    continue
 
-            job.cleanup_cancel_attempted = True
-            try:
-                cancelled = await self.scheduler.cancel_job_async(job.job_id)
-            except Exception as e:
-                logger.warning(
-                    f"Error cancelling active job {job.job_id} "
-                    f"(gen {job.generation}) during cleanup: {e}"
-                )
-                surviving_jobs.append(job)
-                continue
+                try:
+                    cleaned = await asyncio.wait_for(
+                        self._try_cleanup_job(job.job_id, job.results_dir),
+                        timeout=remaining,
+                    )
+                except asyncio.TimeoutError:
+                    cleaned = False
 
-            if cancelled:
-                self.submitted_jobs.pop(str(job.job_id), None)
-                await self._release_evaluation_slot_once(job)
-                cancelled_count += 1
-            else:
-                logger.warning(
-                    f"Failed to cancel active job {job.job_id} "
-                    f"(gen {job.generation}) during cleanup; keeping it tracked"
-                )
-                surviving_jobs.append(job)
+                if cleaned:
+                    self.submitted_jobs.pop(str(job.job_id), None)
+                    await self._release_evaluation_slot_once(job)
+                    cleaned_count += 1
+                else:
+                    surviving_jobs.append(job)
 
-        self.running_jobs = surviving_jobs
-        if cancelled_count:
+            pending_jobs = surviving_jobs
+            remaining = deadline - time.monotonic()
+            if not pending_jobs or remaining <= 0:
+                break
+            await asyncio.sleep(min(retry_delay, remaining))
+            retry_delay = min(_CLEANUP_RETRY_MAX_DELAY_SECONDS, retry_delay * 2)
+
+        self.running_jobs = pending_jobs
+        for job in pending_jobs:
+            logger.error(
+                f"Could not confirm cleanup of active job {job.job_id} "
+                f"(gen {job.generation}) before the cleanup deadline; retaining "
+                "its evaluation slot"
+            )
+        if cleaned_count:
             logger.info(
-                f"Cancelled {cancelled_count} active evaluation job(s) during cleanup"
+                f"Cleaned up {cleaned_count} active evaluation job(s) during shutdown"
             )
 
     async def _cleanup_async(self):
@@ -5611,11 +5648,39 @@ class ShinkaEvolveRunner:
                 getattr(self, "_submission_cleanup_tasks", set())
             )
             if submission_cleanup_tasks:
-                await asyncio.gather(
-                    *submission_cleanup_tasks, return_exceptions=True
+                done, pending = await asyncio.wait(
+                    submission_cleanup_tasks,
+                    timeout=self._get_cleanup_retry_timeout(),
                 )
+                if pending:
+                    for task in pending:
+                        task.cancel()
+                    await asyncio.gather(*pending, return_exceptions=True)
+                    late_job_ids = [
+                        str(job_id)
+                        for job_id in getattr(
+                            self, "_late_submission_jobs", {}
+                        ).values()
+                    ]
+                    logger.error(
+                        "Timed out waiting for late-submission cleanup; surviving "
+                        f"job IDs: {late_job_ids or ['unknown (submission pending)']}"
+                    )
+                await asyncio.gather(*done, return_exceptions=True)
 
             await self._cancel_running_jobs_for_cleanup()
+
+            surviving_job_ids = [
+                str(job.job_id) for job in self.running_jobs
+            ] + [
+                str(job_id)
+                for job_id in getattr(self, "_late_submission_jobs", {}).values()
+            ]
+            if surviving_job_ids:
+                logger.error(
+                    "Evaluation jobs still owned after cleanup deadline: "
+                    f"{surviving_job_ids}"
+                )
 
             # Final recomputation of prompt percentiles to ensure fitness is accurate
             if self.prompt_db is not None and self.db is not None:
--- a/shinka/core/async_runner.py
+++ b/shinka/core/async_runner.py
@@ -535,6 +535,7 @@ class ShinkaEvolveRunner:
         self.finalization_complete = asyncio.Event()
         self.proposal_queue = asyncio.Queue()
         self.active_proposal_tasks: Dict[str, asyncio.Task] = {}
+        self._submission_cleanup_tasks: Set[asyncio.Task] = set()
 
         # Performance tracking
         self.total_proposals_generated = 0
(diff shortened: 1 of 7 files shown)


shinka/launch/local.py
```diff
@@ -35,6 +38,30 @@ def __repr__(self):
         """Return a detailed string representation."""
         return f"ProcessWithLogging(PID: {self.process.pid}, returncode: {self.process.returncode})"
 
+    def _signal_process_group(self, process_group: int, sig: int) -> None:
+        if self.process.poll() is not None:
```
RobertTLange: **Blocking:** the direct child exiting does not mean its process group is empty. A leader can spawn a same-PGID descendant and exit; this guard then makes every later `kill()`/`terminate()` a no-op. I reproduced the grandchild surviving with its PGID still equal to the exited leader's PID. Please use durable group ownership/supervision and add a regression where the leader exits before cancellation.

shinka/core/async_runner.py
```diff
@@ -5461,6 +5500,8 @@ async def _cleanup_async(self):
                     *self.active_proposal_tasks.values(), return_exceptions=True
                 )
 
+            await self._cancel_running_jobs_for_cleanup()
```
RobertTLange: **Blocking:** this only cancels jobs already present in `running_jobs`. If a proposal is cancelled while awaiting executor-backed submission, the executor can still create the process afterward, but the cancelled coroutine never registers it. `CancelledError` also bypasses the evaluation-slot release. I reproduced a live process spawned after cancellation with no slot release. Please track/drain submission futures and cancel/reap any late result; add a blocked-submission regression test.

RobertTLange (changes requested):
The intended process-tree cleanup works for the covered case, and the full gate passes, but two interruption paths still leak evaluators. Requesting changes for the leader-exits-first and cancellation-during-submission races.

shinka/launch/local.py
dexhunter: Fixed in 18d1500 on the rebased head. POSIX evaluators now run beneath a live supervisor that owns the process group after the evaluator leader exits; completion is reported through a private pipe, SIGTERM keeps ownership for escalation, and SIGKILL relinquishes the PGID. The leader-exits-first and TERM-resistant escalation regressions pass. Current-base paired probe: 1 surviving descendant on main, 0 on this branch. Validation: 50 focused tests on Python 3.11 and 717 full-suite tests passed.

shinka/core/async_runner.py
dexhunter: Fixed in 18d1500 on the rebased head. Executor-backed submission is shielded and any late result moves into a runner-owned cleanup task that survives repeated caller cancellation. The evaluation slot remains held until cancellation succeeds or a completed job is status-checked and reaped through the new raw-job-ID scheduler API. Current-base paired probe changes late_jobs_cancelled/slots_in_use from 0/1 to 1/0. Repeated-cancel, failed-cancel, completed-late-job, real-scheduler, and shutdown regressions pass; 50 focused and 717 full-suite tests passed.

dexhunter:
Both blocking findings are addressed on the current head `18d1500`:

- **Leader exits first:** POSIX evaluators now run under a live supervisor that owns the process group after the evaluator exits. Cleanup can still signal the same group, and SIGTERM can escalate to SIGKILL without retaining a reusable PGID.
- **Cancellation during submission:** executor-backed submission is shielded from caller cancellation. A late job is drained by a runner-owned cleanup task, and its evaluation slot stays reserved until the job is cancelled or reaped.

Validation on the current head:

- 50 focused process-cleanup and async-runner recovery tests passed
- 717 full-suite tests passed
- paired leader-exit probe: 1 surviving descendant on main, 0 on this head
- paired late-submission probe: `late_jobs_cancelled/slots_in_use` changed from `0/1` on main to `1/0` on this head

Please re-review the current head when convenient.

dexhunter:
The two requested race fixes are on the current head `18d1500` and CI is green.

- Leader-exits-first cleanup: paired probe reports 0 surviving descendants.
- Cancellation during submission: paired probe reports `late_jobs_cancelled/slots_in_use = 1/0`.
- Validation: 50 focused tests and 717 full-suite tests pass.

Could you please re-review the current head when convenient?

RobertTLange (changes requested):
Thanks, Dex, for the thorough follow-up and regression coverage. The original leader-exits-first and cancellation-during-submission races now appear resolved, and the full gate is green.

Before merging, please address these targeted cleanup and lifecycle gaps:

1. **Make `ProcessWithLogging.send_signal()` group-aware and handle abnormal supervisor exit safely** (`shinka/launch/local.py:53`, `:158`). Delegation currently signals only the supervisor: `SIGTERM` is swallowed, while `SIGKILL` kills the supervisor but leaves the evaluator alive. `cleanup_logging()` then skips group cleanup because the supervisor has exited. Please route signals to the owned group and ensure cleanup cannot retain or later signal a stale PGID. Add regressions for `send_signal(SIGTERM)` and `send_signal(SIGKILL)`.

2. **Confirm Slurm termination before releasing the evaluation slot** (`shinka/core/async_runner.py:5089`, `shinka/launch/scheduler.py:461`). A zero `scancel` exit code confirms command acceptance, not that the job has left the queue. Please poll to terminal disappearance or reap before releasing the slot, with a delayed-cancellation regression.

3. **Bound late-submission cleanup during shutdown** (`shinka/core/async_runner.py:5079`, `:5610`). The retry loop is infinite and `_cleanup_async()` gathers it without a deadline, so a permanent scheduler or status failure blocks database close and scheduler shutdown forever. Please add a bounded cleanup deadline or escalation policy, report any surviving job IDs explicitly, and cover the permanently failing case.

4. **Retry transient failures when cancelling already-running jobs** (`shinka/core/async_runner.py:5568`). `cleanup_cancel_attempted` is set before success; failed jobs are then skipped forever even though the scheduler is immediately shut down. Please retry to a bounded deadline and confirm cancellation or terminal state before considering cleanup complete. Add a transient-failure regression.

5. **Keep completion-pipe descriptors away from standard FDs** (`shinka/launch/local.py:240`). If stdin and stdout are closed, `os.pipe()` can return FDs 0 and 1; `stdout=PIPE` then replaces the completion writer. In a focused repro, the return code was written to `job_log.out` and `wait()` timed out. Please relocate both pipe ends to FDs >= 3 with close-on-exec semantics and add a closed-standard-FD harness test.

Once these are addressed and covered by focused regressions, this should be ready for another pass. Thank you again for the careful work here.

dexhunter:
All requested lifecycle fixes are now addressed on `f93bbd51`:

- abnormal supervisor EOF remains armed independently of the completion FD; the `None, None, -SIGKILL` race now routes `poll`, `returncode`, `wait`, and cleanup through fail-closed handling exactly once, kills the evaluator group, and clears PGID/pending state without a later stale signal
- `SIGTERM`/`SIGKILL` target the owned process group
- Slurm cancellation waits for queue disappearance before releasing a slot; status errors remain unknown/nonterminal
- late submissions and active cancellation retries share a bounded shutdown deadline; unresolved ownership is retained and logged
- blocked `scancel`/`squeue` timeout and error paths cannot block executor shutdown
- completion-pipe descriptors are relocated above standard FDs with close-on-exec semantics

Validation on the exact head: 70 focused lifecycle tests, 736 full-suite tests (1 deselected), mypy, Ruff, and GitHub CI pass. An independent 300-iteration abnormal-exit stress run without direct `process.wait()` had zero leaks or stale-PGID signals.

dexhunter:
@RobertTLange the requested cleanup follow-ups are complete on the current head, with the validation summary above. Could you re-review when convenient?

dexhunter:
Addressed the remaining late-submission lifecycle races in `79d1aa6`:

- the scheduler now retains ownership until a submission is either claimed or cleanup is durably queued
- failed cancellation uses bounded retries and does not release the evaluation slot prematurely
- concurrent shutdown cannot close the executor before cleanup work is queued
- added deterministic coverage for late completion, cancellation failure, and concurrent shutdown

Validation on the final patch:

- focused lifecycle/scheduler tests: `59 passed`
- full suite: `739 passed, 1 deselected`
- Ruff, test-file mypy, and `git diff --check`: passed
- current-upstream merge/apply check: passed

CI is also green on the updated head.

dexhunter:
A quick status note so the review state doesn't mislead: GitHub still shows **changes requested**, but each of the five items from the 2026-07-15 review is implemented on the current head `79d1aa6`. Per item, with where to verify:

1. **Group-aware `send_signal()` with safe handling of an abnormal supervisor exit** — signals route through `_signal_process_group` whenever a group is owned (`local.py:185-191`); regressions: `test_send_signal_sigterm_reaches_process_group`, `test_send_signal_sigkill_reaches_process_group`.
2. **Confirm Slurm termination before releasing the slot** — the `scancel` return code is not trusted; `scheduler.py:576-614` polls `get_job_status` until the job has left the queue under a bounded deadline and returns `False` on timeout; regression: `test_cleanup_keeps_slot_until_cancellation_is_confirmed`.
3. **Bound late-submission cleanup during shutdown** — `async_runner.py:5193-5223` applies `_get_cleanup_retry_timeout` with a monotonic deadline and logs when a submission retains its evaluation slot; regression: `test_cleanup_deadline_bounds_blocked_slurm_command`.
4. **Retry transient failures while cancelling running jobs** — bounded-deadline retry path; regression: `test_cleanup_retries_transient_active_job_cancellation`.
5. **Keep completion-pipe FDs off the standard streams** — `_move_fd_above_standard_streams` is applied to both pipe ends before `Popen` (`local.py:291-299`); regression: `test_completion_pipe_avoids_closed_standard_fds`.

The two remaining review threads point at code that has since changed (GitHub marks them outdated). As of this comment the PR is mergeable against `main` with CI green. Happy to make any further change you'd like before re-review.

dexhunter:
Re-verified against today's `main` (`b67a073`), since the last CI run here was 2026-07-17 and eight commits have landed since:

- merges cleanly, no conflicts
- full suite on the merged tree: `758 passed, 2 deselected`; the `mypy` CI step is clean; the 73 tests in the three files this PR adds all pass
- the six regressions named in my note above pass on `79d1aa6` under Python 3.11, the CI interpreter

One heads-up so a re-run here does not mislead: CI in this repo currently fails at the Ruff step on every branch, including unmodified `main`. `ruff` is unpinned in `[dependency-groups] dev`, so `uv sync --dev` moved from 0.15.22 (last green run, 2026-07-23) to 0.16.0, whose wider default rule set reports 153 errors across 36 test files. I wrote up the breakdown in #168. Nothing to do on this PR — I just did not want a red check here to read as a regression from it.

dexhunter:
I see `fix/correctness-concurrency` (#150) now rebuilds this surface from the same base as this PR, with a durable token identity, a forked group guard and `PR_SET_PDEATHSIG`. That is a better foundation than what I built here, and reading it, it independently covers four of the five items you left on this PR:

1. `send_signal`/`terminate`/`kill` route through the identity rather than the supervisor handle (`local.py:290-330`), covered by `test_kill_terminates_child_process_group`, `test_supervisor_termination_reaps_sanitized_evaluator`, `test_supervisor_sigkill_reaps_wrapper_grandchild`.
2. Slot release waits for queue disappearance — `test_scheduler_retains_known_job_id_until_it_disappears`, `test_cleanup_waits_for_known_job_id_to_disappear`.
3. Late-submission cleanup retains and reports rather than looping — `test_cleanup_retains_late_submission_until_cancellation_is_confirmed`, `test_parent_death_exec_prevents_late_submission_side_effect`.
4. Transient cancellation failures retry — `test_cleanup_retries_and_retains_job_when_cancellation_fails`, `test_cleanup_releases_job_after_retry_succeeds`, `test_cleanup_retries_cancellation_exceptions`.

Given that, this PR is in your way more than it is helping. I'm happy for it to be closed in favour of #150 — just say the word and I'll close it.

**One item does not carry over, and it is item 5 from your review of this PR.**

At `local.py:458` on that branch the status pipe is used at whatever descriptors `os.pipe()` hands back, and the write end's *number* is passed to the supervisor as `argv[1]` while the same `Popen` sets `stdout=subprocess.PIPE`. If the parent's standard streams are closed, `os.pipe()` returns `(0, 1)`, `subprocess` dup2s the stdout pipe onto FD 1 in the child, and the supervisor's `READY` goes to stdout instead of the status pipe.

I ran that branch's own `_local_supervisor.py` under the launch sequence from `local.py:454-546`. Three arms, only the descriptors differ:

| parent std fds | `os.pipe()` returns | status pipe | `job_log.out` | launch outcome |
| --- | --- | --- | --- | --- |
| open | `(3, 4)` | `READY` | empty | ok |
| 0 and 1 closed | `(0, 1)` | EOF, 0 bytes | `READY` | `RuntimeError: Local evaluator supervisor exited before launch` |
| 0 and 1 closed, both ends moved to >= 3 | `(0, 1)` -> `(3, 4)` | `READY` | empty | ok |

So the supervisor starts fine and reports correctly; the parent just cannot hear it, takes the `launch_status != "READY"` branch at `local.py:546`, and then kills the evaluator it had successfully launched. Same shape as the repro you described in item 5 here — the message ends up in `job_log.out` and the parent concludes the launch failed — now on the new status pipe rather than the old completion pipe.

The fix is one call on each end before `Popen` — `fcntl(fd, F_DUPFD_CLOEXEC, 3)`, with `pass_fds` clearing CLOEXEC again for the write end in the child, so nothing else changes. That is exactly what `_move_fd_above_standard_streams` does on this branch (`local.py:34`); the third arm above is that helper, byte for byte, dropped into the #150 sequence with nothing else altered.

Happy to send it as a small standalone PR against `main` so it does not depend on either of these branches — whichever is least disruptive for you.

dexhunter:
Closing this in favour of #150 — that branch rebuilds the same surface on a better foundation, and it should be yours to carry rather than something you have to reconcile against an outside PR.

One thing from here that I could not find in #150, handing it over rather than leaving it buried in a closed PR: the status pipe at `local.py:458` is used at whatever descriptors `os.pipe()` returns, while the same `Popen` sets `stdout=subprocess.PIPE`. With the parent's standard streams closed, `os.pipe()` returns `(0, 1)` and the supervisor's `READY` goes to stdout instead of the status pipe — the three-arm repro is in my previous comment.

The whole fix is `fcntl(fd, F_DUPFD_CLOEXEC, 3)` on both ends before `Popen`, with `pass_fds` clearing CLOEXEC again for the write end in the child. Two things here are free to take verbatim, no attribution needed:

- `_move_fd_above_standard_streams` (`shinka/launch/local.py:34`), applied at both pipe ends
- `test_completion_pipe_avoids_closed_standard_fds` (`tests/test_local_process_cleanup.py:410`), the closed-standard-FD harness test you asked for

Thanks for the reviews on this one — they were detailed and genuinely useful, and #150 is clearly the better home for the work.

PR #149, opened 2026-07-16
Add optional Weights & Biases logging

anantgar:
## Summary

- add optional Weights & Biases logging alongside the existing SQLite and WebUI paths
- log every evaluated individual's score against its generation, plus compact evaluation, timing, and cost metrics

## Why

For researchers who prefer metrics reporting online on Weights & Biases.

## Configuration and behavior

W&B logging is disabled by default and is enabled with `evo.enable_wandb_logging=true`. The integration is best-effort: W&B failures emit warnings without interrupting evolution or replacing existing logging.

Online runs use `wandb login` or `WANDB_API_KEY`. The secret-backed integration workflow installs the optional `wandb` extra and expects the SakanaAI repository to configure a `WANDB_API_KEY` Actions secret.

## Testing

- `uv run --with pytest-cov pytest -q -m "not requires_secrets" --cov=shinka --cov-report=term-missing --cov-report=xml:coverage.xml` — 724 passed, 1 deselected
- `uv run pytest -q -m "requires_secrets"` — 1 passed, 724 deselected against the authenticated W&B service

## Compatibility

- no changes to SQLite or WebUI logging
- no W&B dependency unless the optional `wandb` extra is installed

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -24,6 +24,7 @@ jobs:
       AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
       AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
       AWS_REGION_NAME: ${{ secrets.AWS_REGION_NAME }}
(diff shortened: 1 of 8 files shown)


RobertTLange:
Thanks @anantgar for contributing the optional Weights & Biases logging integration!

PR #150, opened 2026-07-17
fix: correctness & concurrency findings from deep review

RobertTLange:
## Correctness & Concurrency fixes

Fixes the Critical/High/Medium correctness and concurrency findings from the deep review (`docs/deepreview.md` §2). Each fix ships with regression tests; the full suite is green (**767 passed**, was 723 — 44 new tests) and ruff/mypy pass.

This is the first of a stacked series of per-category PRs; later category branches build on this one.

### Critical — silent result corruption in the diff engine (`shinka/edit/apply_diff.py`)
- **Substring matching** — `SEARCH "xval = 10"` matched inside `maxval = 100` via `str.find` and rewrote it to `maxval = 9990`, reporting success. Matching is now **line-aligned** (leading-indent offsets still allowed) and applied at the validated position instead of `str.replace`, which could land outside the checked editable region. *(reproduced by execution)*
- **Dropped deletion hunks** — an empty-REPLACE (deletion) hunk failed to parse and was silently dropped while the multi-hunk patch still reported success. `PATCH_PATTERN` now accepts an empty REPLACE, and a `<<<<<<< SEARCH` header that produces no parsed block raises instead of reporting partial success. *(reproduced by execution)*
- Also: prefer an editable-region occurrence over an identical line in an earlier immutable region; reject ambiguous (non-unique) SEARCH targets; stop dropping the file's trailing newline / splitting on form-feeds.

### High
- **Eval-result crash consistency** (`wrap_eval.py`, `utils/general.py`) — write `metrics.json` before `correct.json` (success marker last, fsync'd) so an interruption between the two writes reads as `correct=False` instead of recording a passing program with a missing score as `correct=True, combined_score=0.0`. Decode `correct.json` defensively. Treat a missing `combined_score` as incorrect.
- **Launch robustness** (`launch/local.py`, `launch/slurm.py`) — kill the whole process group on timeout/cancel (`start_new_session` + `killpg`) so wrapper commands (`conda run`, `bash -lc … docker run`) don't orphan GPU-holding children; a local job still waiting for a GPU reports as running (not done); a `squeue` failure for a departed job is resolved via `sacct` instead of hanging as "running forever"; the sync monitor bounds unknown statuses.
- **Async job-monitor race** (`core/async_runner.py`) — rebuild `running_jobs` from a fresh read minus jobs just resolved as completed, instead of a snapshot taken before the `cancel_job_async` await. The stale snapshot silently dropped jobs a proposal task appended during that await, leaking evaluation slots (eventual pool deadlock).
- **Resume detection** — gate on persisted program count, not `last_iteration > 0`; a run that persisted only the generation-0 cohort before dying was mis-detected as a fresh start (duplicate cohort, skipped bandit/cost restore, inflated generation count → premature stop).
- **Signal shutdown** — SIGINT/SIGTERM now request a graceful stop and every running eval job is cancelled during cleanup, so Ctrl-C no longer orphans local subprocesses or leaves Slurm jobs on the cluster.
- **Empty/truncated LLM output** (`providers/gemini.py`, `providers/local_openai.py`) — raise on empty content and truncation (`MAX_TOKENS`/`length`) instead of returning `content=""` as a finished program.

### Medium
- `extract_between` returns `None` (not the truthy `"none"`) on no-match; callers updated so a failed extraction is detected instead of injecting the literal string as code.
- Novelty judge fails **closed** on an empty/blocked response (reject rather than silently accept), while a transient API exception stays fail-open.
- `QueryResult.__str__` guards the thinking/output token ratio against `output_tokens == 0` (crashed logging of exactly the degenerate responses worth inspecting).
- `get_best_program` no longer treats an exactly-`0.0` score as `-inf` (falsy-zero); island samplers use a stable softmax / non-negative monotonic weights with uniform fallback (no more NaN crashes or inverted preference); `AsymmetricUCB._posterior_batch` no longer clobbers its loop counter with the cost coefficient.

### Deferred (documented, not in this PR)
- **L1 — non-atomic best-program / archive / migration read-modify-write.** The reviewer verified this is **not exploitable today** (all DB maintenance runs serially in one coroutine with a `max_workers=1` write executor; the background-worker pool that would parallelize it has zero call sites). A correct fix needs a `BEGIN IMMEDIATE` transaction plus in-memory-cache invalidation and should land alongside any revival of that concurrency subsystem — making untested transaction-boundary changes to the hot DB path now would risk lock contention/deadlocks. Tracked for that future work.
- The subtle `apply_diff` indentation double-indent behavior (Medium) is left as-is to avoid regressing the indentation-correction tests that pin current behavior; the higher-value line-anchoring fix already narrows its surface.

### Testing
`uv run pytest -q -m "not requires_secrets and not models_dev_live"` → 767 passed. `ruff check tests` and `mypy` (CI config) pass. New regression tests: diff substring-collision & empty-REPLACE-deletion, partial-write `correct.json`/`metrics.json`, missing `combined_score`, job-status/process-group teardown, graceful-shutdown job cancellation, empty/truncated provider output, token-ratio divide-by-zero, falsy-zero best-program, island-sampler overflow/negative-fitness.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,11 @@ source .venv/bin/activate  # On Windows: .venv\Scripts\activate
 uv pip install -e .
 ```
 
+> [!NOTE]
+> The async runner's secure results-directory and durable-resume workflow
+> currently requires Linux. It fails closed on macOS and Windows because
+> equivalent native ACL and reparse-point validation is not yet implemented.
+
 For detailed installation instructions and usage examples, see the [Getting Started Guide](https://sakanaai.github.io/ShinkaEvolve/getting_started/).
 
 ## Examples 📖
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ uv pip install -e .
 ```
 
 > [!NOTE]
-> The async runner's secure results-directory and durable-resume workflow
+> The async runner's secure results-directory, durable-resume, and local-job workflow
 > currently requires Linux. It fails closed on macOS and Windows because
 > equivalent native ACL and reparse-point validation is not yet implemented.
 
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -57,7 +57,7 @@ uv pip install -e .
 ```
 
 !!! note
-    The async runner's secure results-directory and durable-resume workflow
+    The async runner's secure results-directory, durable-resume, and local-job workflow
     currently requires Linux. It fails closed on macOS and Windows because
     equivalent native ACL and reparse-point validation is not yet implemented.
 
--- a/docs/getting_started.md
+++ b/docs/getting_started.md
@@ -56,6 +56,11 @@ source .venv/bin/activate    # macOS/Linux
 uv pip install -e .
 ```
 
+!!! note
+    The async runner's secure results-directory and durable-resume workflow
+    currently requires Linux. It fails closed on macOS and Windows because
+    equivalent native ACL and reparse-point validation is not yet implemented.
+
 ### Option 2: conda/pip
 
 ```bash
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -86,7 +86,7 @@ shinka = [
 dev = [
     "pytest>=6.0",
     "pytest-cov",
-    "ruff",
+    "ruff==0.15.20",
     "mypy",
     "black",
     "isort",
--- a/shinka/core/async_novelty_judge.py
+++ b/shinka/core/async_novelty_judge.py
@@ -217,8 +217,12 @@ class AsyncNoveltyJudge:
             )
 
             if response is None or response.content is None:
-                logger.warning("Novelty LLM returned empty response")
-                return True, "LLM response was empty", 0.0
+                # Fail CLOSED: a transient outage/empty response must not be
+                # silently accepted as novel (is_novel=False => reject upstream).
+                logger.warning(
+                    "Novelty LLM returned empty response; rejecting as not novel"
+                )
+                return False, "LLM response was empty (failing closed)", 0.0
 
             content = response.content.strip()
             api_cost = response.cost or 0.0
@@ -255,7 +259,8 @@ class AsyncNoveltyJudge:
             )
 
             if not response or not response.content:
-                return True, 0.0, "No response from LLM"
+                # Fail CLOSED on empty response (is_novel=False => reject).
+                return False, 0.0, "No response from LLM (failing closed)"
 
             # Parse response for novelty decision
             is_novel = self._parse_novelty_response(response.content)
--- a/shinka/core/async_novelty_judge.py
+++ b/shinka/core/async_novelty_judge.py
@@ -222,7 +222,8 @@ class AsyncNoveltyJudge:
                 logger.warning(
                     "Novelty LLM returned empty response; rejecting as not novel"
                 )
-                return False, "LLM response was empty (failing closed)", 0.0
+                cost = response.cost if response is not None else 0.0
+                return False, "LLM response was empty (failing closed)", cost
 
             content = response.content.strip()
             api_cost = response.cost or 0.0
@@ -260,7 +261,8 @@ class AsyncNoveltyJudge:
 
             if not response or not response.content:
                 # Fail CLOSED on empty response (is_novel=False => reject).
-                return False, 0.0, "No response from LLM (failing closed)"
+                cost = response.cost if response is not None else 0.0
+                return False, cost, "No response from LLM (failing closed)"
 
             # Parse response for novelty decision
             is_novel = self._parse_novelty_response(response.content)
@@ -289,4 +291,3 @@ class AsyncNoveltyJudge:
     def __getattr__(self, name):
         """Delegate unknown methods to sync novelty judge."""
         return getattr(self.sync_judge, name)
-
--- a/shinka/core/async_novelty_judge.py
+++ b/shinka/core/async_novelty_judge.py
@@ -211,14 +211,19 @@ class AsyncNoveltyJudge:
         )
 
         try:
-            response = await self.async_llm_client.query(
+            query_method = getattr(
+                self.async_llm_client,
+                "query_or_raise",
+                self.async_llm_client.query,
+            )
+            response = await query_method(
                 msg=user_msg,
                 system_msg=NOVELTY_SYSTEM_MSG,
             )
 
             if response is None or response.content is None:
-                # Fail CLOSED: a transient outage/empty response must not be
-                # silently accepted as novel (is_novel=False => reject upstream).
+                # Fail closed only for a completed query without a usable verdict.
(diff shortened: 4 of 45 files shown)


RobertTLange:
## Review follow-up

Fixed each review finding in an isolated commit:

- `92c8a6c` — require terminal scheduler evidence before treating Slurm jobs as complete
- `c070a55` — reject unterminated SEARCH/REPLACE hunks
- `844215c` — preserve evaluation-job ownership during shutdown
- `de49875` — retain ownership when cancellation cannot be confirmed
- `d06600c` — make signal-triggered shutdown cancellation-safe
- `b864949` — stop retrying rejected Gemini responses while retaining billed cost/result position
- `70c6b0b` — correctly resume generation-zero prompt archives
- `3cf2fe1` — retain ambiguous Slurm submissions by unique name through cleanup
- `baf9cf3` — checkpoint billed proposal, embedding, novelty, prompt, and meta costs; prevent double counting and fail closed on invalid costs
- `aeab3d2` — confirm known-ID Slurm cancellation before releasing ownership, with a bounded ~31s terminal window
- `c756d37` — checkpoint multi-step and partial-batch meta responses incrementally without aggregate double counting

Validation:

- `847 passed, 2 deselected` (`requires_secrets` and `models_dev_live` excluded); 5 existing NumPy bandit warnings
- Ruff passed for the test suite (excluding existing `tests/file.py` debt)
- Mypy passed for all 71 test source files
- Final aggregate correctness, reliability, and security review: no remaining actionable findings
- Worktree clean; all commits pushed

The PR had no existing review threads to reply to or resolve, so this comment records the findings and fixes.

RobertTLange:
## Review findings addressed

1. **Stale correctness markers could be reused** — invalidated markers before evaluation and made result publication durable. Fixed in `945bb87`.
2. **Novelty provider outages looked like valid empty responses** — separated provider failure from completed empty output, with fail-open outage behavior and fail-closed empty-response behavior. Fixed in `da5e137`.
3. **Unknown Slurm states could poll forever** — preserved unknown state, bounded the wait, and made timeout cleanup fatal/observable. Fixed in `be1867d`.
4. **Resume replayed/archived the wrong generation state** — resume now derives exact missing generations and archives interrupted directories safely. Fixed in `23fdc71`.
5. **Async execution replaced host signal handlers** — direct runs are signal-neutral; the wrapper installs and restores only handlers it owns. Fixed in `fea819e`.
6. **Evaluation ownership existed only in memory** — added durable SQLite ownership intent/activation/resolution, startup reconciliation, identity-bound Slurm recovery, post-launch orphan cleanup, fail-fast ownership conflicts, and lifecycle coverage for initial/evolved jobs. Fixed in `c9da7d0`.
7. **Unbounded Ruff resolution broke CI when 0.16.0 landed** — pinned the known-clean dev-only Ruff 0.15.20 toolchain for reproducible unlocked CI. Fixed in `39460ab`.

## Validation

- Focused recovery/concurrency suites: **193 passed**
- Full non-secret/non-live suite: **915 passed, 2 deselected**
- Ruff CI command: passed in a fresh isolated resolution
- mypy CI command: **71 source files**, passed
- MkDocs strict build: passed
- Final regression, reliability, and security re-reviews: no material findings
- Push and pull-request CI runs for `39460ab`: passed

RobertTLange:
## Review fixes

Addressed each correctness/reliability finding in a separate commit:

1. `dcb4768` — propagate database count failures instead of treating an existing run as fresh.
2. `6c83bf8` — persist and re-authenticate local process ownership so jobs can be safely recovered/cancelled after a controller crash.
3. `3618b43` — resume from the number of distinct persisted generation IDs, not the program row count.
4. `370f0f1` — search all indentation-equivalent matches before declaring a diff ambiguous or missing.
5. `dd355ca` — accept group-writable resume roots only when owned by the current user and shared with one of the process's groups; world-writable roots remain rejected.
6. `e0d5c58` — recover timed-out `sbatch` submissions from both `squeue` and recent `sacct` accounting data, with UID/name/job-ID validation and bounded polling.
7. `ea40629` — align existing Slurm recovery fixtures with the strengthened queue/accounting protocol.
8. `b979fba` — preserve relative dedentation when applying nested indentation-equivalent replacements (follow-up found by the final review pass).

## Validation

- Per-finding regression tests: red before each fix, green afterward.
- Final suite: **940 passed, 2 deselected**.
- Ruff: passed.
- mypy: passed.
- MkDocs strict build: passed.
- sdist/wheel build: passed.
- `uv lock --check`: passed.
- Final parallel regression, reliability, and security reviews: no remaining actionable findings.
- Bandit on touched production files: no medium/high findings. Full-repository Bandit and dependency audit still report pre-existing baseline debt (including the existing Pillow 11.3 advisories); this fix series does not modify those dependencies.

There were no existing inline review threads to reply to or resolve.

RobertTLange:
## Follow-up review fixes

Reviewed the post-`b979fba` state and addressed each new finding in a separate commit:

1. `84b382c` — persist submission identity before launch, closing the dispatch-before-ownership crash window.
2. `eaf108c` — validate the results root before startup side effects.
3. `79e41c1` — reject unsafe results-root ACLs in addition to unsafe mode bits.
4. `023b6b4` — retain authenticated local ownership when evaluator environments are sanitized.
5. `477c7ca` — wait for the authenticated local process group, not only its leader, before reporting completion.
6. `f672748` — preserve replacement indentation when matched context starts with blank lines.
7. `71439ef` — bind Slurm recovery results to the current user and allocation.
8. `0310fdd` — add a parent-death guard so supervisor loss cannot orphan the local evaluator group.
9. `049e26a` — distinguish recovered, absent, and unavailable Slurm submission outcomes.
10. `5af0465` — retain durable ownership when Slurm cleanup is ambiguous.
11. `a29cffc` — hold an exclusive results-root lease for the runner lifetime.
12. `7f6b67e` — bridge that lease across local fork/exec until the durable process token is visible, with allowlisted descriptors, a minimal trusted handoff environment, exact-inode Bash validation, and authenticated post-spawn cleanup.

## Validation

- Per-finding regressions were validated before each isolated commit.
- Full suite: **1,013 passed, 2 deselected** (`requires_secrets` and `models_dev_live`).
- Ruff CI command: passed.
- Mypy CI command: passed for **71 source files**.
- MkDocs strict build: passed.
- sdist/wheel build: passed; all local launch helpers are packaged.
- `uv lock --check`: passed.
- Bandit: no high-severity findings; two pre-existing low-confidence `B608` warnings in `shinka/database/dbase.py` outside this fix range.
- Final aggregate correctness, reliability, and security reviews: no remaining actionable findings.
- No inline review threads existed to reply to or resolve.

RobertTLange:
Thanks for the deep investigation and extensive regression work. We are changing the integration policy: this multi-subsystem parent is too broad to merge as one unit. Please revive high-confidence fixes as small, current-main PRs; async/scheduling changes additionally need the Circle Packing parallelism matrix.

PR #151, opened 2026-07-17
fix: security hardening from deep review (server, XSS, launch)

RobertTLange:
## Security hardening from deep review

Fixes the High/Medium security findings from the deep review (`docs/deepreview.md` §1), concentrated in the visualization server and web UI. New regression tests; full suite green; ruff/mypy pass.

**Stacked on `fix/correctness-concurrency`** (base branch of this PR) — review/merge that first.

### Visualization server (`shinka/webui/visualization.py`)
- **Bind 127.0.0.1 by default** (was `0.0.0.0`, network-reachable with no auth). External exposure is now an explicit `--host` opt-in that prints a no-auth warning. Threaded server so one slow request (PDF export) can't freeze the UI.
- **Path traversal** — every user-supplied `db_path`/meta/plot path is now validated against the search root with `realpath` + `commonpath` (rejecting absolute paths, `..`, and in-root symlinks that escape). Previously `_get_actual_db_path` was a no-op, so `?db_path=/abs/x.db` opened any SQLite DB or file on the host.
- **/plot_file/** containment-checks before touching disk (no 403-vs-404 existence leak) and restricts to image extensions.
- **CORS** — dropped the `Access-Control-Allow-Origin: *` wildcard (UI is same-origin) and added a **Host-header allowlist** so a DNS-rebinding page can't read the DB. `X-Content-Type-Options: nosniff` on all responses; directory listings disabled.
- **PDF export** — pass `--disable-javascript --disable-local-file-access --disable-external-links` to wkhtmltopdf, which renders LLM-generated meta HTML and would otherwise run JS / load `file://` (local-file exfiltration / SSRF).

### Web UI XSS + supply chain (`shinka/webui/viz_tree.html`)
- **Stored XSS** — DB content (patch names, thoughts, meta text, prompt descriptions) is attacker-influenced (LLM output). Consolidated the three divergent `escapeHtml` definitions into one that also escapes quotes (the live copy didn't), routed all markdown through `DOMPurify.sanitize(marked.parse(...))` with an escaped-plain-text fallback, and escaped the remaining raw DB-field `innerHTML`/d3-`.html()` interpolations.
- **SRI** — pinned every CDN script to an exact version with a Subresource Integrity hash + `crossorigin` (marked and plotly were unpinned), and added DOMPurify. Verified: the extracted inline script passes `node --check`.

### Launch (`shinka/launch/`)
- `shlex.quote` env values spliced into Slurm/conda/docker shell scripts (an `eval_env` value with shell metacharacters could otherwise inject commands).
- **Opt-in eval-secret stripping** — `SHINKA_STRIP_EVAL_SECRETS=1` (default off) withholds provider credentials from the local eval subprocess, which runs LLM-generated code unsandboxed and inherits the parent environment. Explicit `env_overrides` are applied after stripping. The code-execution trust boundary is now documented on `load_program`.

### Config (`shinka/env.py`)
- The package-dir `.env` is loaded with `override=False` so a stray/shipped `.env` can't silently shadow the operator's real environment; the launch-dir `.env` keeps its intended precedence.

### Not changed (documented recommendations)
- **S3 full sandboxing** — local eval remains unsandboxed by design (the framework runs evolved code). Beyond the opt-in secret stripping and docs added here, running untrusted models should use the Docker/Slurm eval path; a full sandbox is out of scope for this PR.
- **`.github/workflows/claude.yml`** triggers `claude-code-action` on `@claude` from any GitHub user (prompt-injection / billing-DoS surface). Its permissions are read-only, limiting blast radius; restricting the trigger to members/authors is a repo-policy decision left to maintainers.

### Testing
New `tests/test_webui_security.py` (path-traversal unit + HTTP integration: absolute/`..`/symlink/sibling rejection, DNS-rebinding Host check, no CORS wildcard, nosniff, disabled listing) and launch/env tests (shell-quoting, secret stripping, package-.env shell precedence). Full suite green; ruff + mypy pass.

### DIFF
--- a/shinka/core/wrap_eval.py
+++ b/shinka/core/wrap_eval.py
@@ -25,7 +25,14 @@ DEFAULT_METRICS_ON_ERROR = {
 
 
 def load_program(program_path: str) -> Any:
-    """Loads a Python module dynamically from a given file path."""
+    """Loads a Python module dynamically from a given file path.
+
+    Security note: ``program_path`` points at LLM-generated code, and importing
+    it executes it in-process with no sandbox. Run evolution only with models
+    and initial programs you trust, or use an isolated eval (Docker/Slurm). The
+    local launcher honours ``SHINKA_STRIP_EVAL_SECRETS=1`` to withhold provider
+    credentials from the eval subprocess (see shinka.launch.local).
+    """
     spec = importlib.util.spec_from_file_location("program", program_path)
     if spec is None:
         raise ImportError(f"Could not load spec for module at {program_path}")
--- a/shinka/env.py
+++ b/shinka/env.py
@@ -23,6 +23,10 @@ def load_shinka_dotenv(
         env_paths.append(launch_env)
 
     for env_path in env_paths:
-        load_dotenv(dotenv_path=env_path, override=True)
+        # The package-dir .env fills gaps only (override=False) so a stray or
+        # shipped .env cannot silently shadow the operator's real environment.
+        # The launch-dir .env keeps its intended precedence (override=True).
+        override = env_path == launch_env
+        load_dotenv(dotenv_path=env_path, override=override)
 
     return tuple(env_paths)
--- a/shinka/launch/local.py
+++ b/shinka/launch/local.py
@@ -10,6 +10,48 @@ import logging
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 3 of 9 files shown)


RobertTLange:
- #157 WebUI server boundary security
- #158 WebUI content and PDF security
- #159 launch environment security

PR #152, opened 2026-07-17
perf: performance fixes from deep review (indexes, N+1, caching)

RobertTLange:
## Performance fixes from deep review

Addresses the High/Medium performance findings from the deep review (`docs/deepreview.md` §3), focused on the per-generation DB hot path and the web-UI server. New regression tests; full suite green; ruff/mypy pass.

**Stacked on `fix/security-hardening`** (base branch of this PR).

### Database
- **Index the two hottest columns** — added composite indexes `(correct, combined_score)` and `(island_idx, correct, combined_score)`. The per-generation top-N / elite / island-aggregate queries filter on `correct` and sort by `combined_score`, previously forcing a full scan + filesort every generation. `EXPLAIN QUERY PLAN` now shows an index range scan. `IF NOT EXISTS`, so existing DBs pick them up on open.
- **N+1 in archive inspiration sampling** — `ArchiveInspirationSelector` fetched candidate ids then issued a `get_program()` per id every generation (each re-parsing code + embeddings). Now one `SELECT p.*` per category, building programs via the already-wired `program_from_row` converter (the pattern top-k sampling already used).
- **Novelty check scanned each island twice** — `compute_similarity` and `get_most_similar_program` each walked the whole island and re-parsed every embedding independently. A shared single-pass `_scan_island_similarities` + `compute_similarity_details` / `compute_similarity_details_thread_safe` return `(scores, most-similar program)` from one scan; both novelty judges route through them (the thread-safe variant fetches the winner in the same connection, so the async path makes one executor call instead of two full scans).

### Web UI server
- **3s change-detection poll made cheap** — `handle_get_program_count` no longer loads and JSON-parses every failed-proposal node and reads a file per node; it returns count + max timestamp + a failed-proposal COUNT via an aggregate query.
- **`/get_programs_summary` now TTL-cached** like `/get_programs`, so auto-refresh during active evolution doesn't re-read/re-parse the whole table each time. Both handlers key the cache on the resolved absolute path (distinct namespaces) — fixing a latent collision where a shared relative `db_path` could serve one handler's payload to the other.
- **`/get_programs` opt-in narrowing** — `limit`/`offset`/`include_code` query params; the default (no params) response is byte-for-byte unchanged so existing `viz_tree.html` keeps working.

### Launch
- **Batch Slurm status into one `squeue` per poll tick** — `batch_check_status_async` resolved each Slurm job with its own `squeue` subprocess every tick (~10·N spawns/sec against slurmctld). It now runs a single `squeue --jobs=id1,id2,...` per tick and falls back to the per-job path (with sacct disambiguation) if the batched query is unavailable. Local jobs keep the cheap `Popen.poll()` path.

### LLM
- **Memoize SDK clients** — the anthropic/openai/google client (and its httpx pool) was rebuilt on every query and retry. Now cached by provider + resolved model + structured-output flag + base URL + a non-reversible fingerprint of the relevant key(s); async clients cached per event loop (WeakKeyDictionary) so a loop-bound client is never reused across loops.

### Deferred (documented recommendations)
- **Embeddings as float32 BLOB (P3)** — high value but a storage-format migration with real backward-compat risk for existing JSON-embedding databases; left as follow-up. The single-pass novelty scan above already removes the redundant work without a format change.
- **Sync `add()` PCA/GMM throttle (P5)** — the production path is async, which already throttles the recompute to every `embedding_recompute_interval` inserts in a background thread; the sync path is not the production hot path and throttling it risks breaking tests that expect immediate PCA columns.
- **Per-call async DB reconstruction / connection pooling (P7)** and the parents.py power-law N+1 need connection-lifecycle / row-converter threading that pairs naturally with the duplication cleanup; deferred there to avoid churn and concurrency risk here.

### Testing
New tests: DB index existence + `EXPLAIN QUERY PLAN` uses the index; inspiration sampling returns fully-hydrated programs; single-pass similarity matches the legacy two-method result; batched `squeue`; web-UI count/summary-cache/narrowing; LLM client cache identity. Full suite green; ruff + mypy (CI config) pass.

### DIFF
--- a/shinka/core/async_novelty_judge.py
+++ b/shinka/core/async_novelty_judge.py
@@ -91,9 +91,11 @@ class AsyncNoveltyJudge:
                 return True, novelty_metadata
 
             loop = asyncio.get_event_loop()
-            similarity_scores = await loop.run_in_executor(
+            # One thread-safe scan returns both the scores and the most similar
+            # program, replacing two full island scans per candidate.
+            similarity_scores, most_similar_program = await loop.run_in_executor(
                 None,
-                db.compute_similarity_thread_safe,
+                db.compute_similarity_details_thread_safe,
                 code_embedding,
                 parent_program.island_idx,
             )
@@ -127,15 +129,7 @@ class AsyncNoveltyJudge:
             novelty_cost = 0.0
 
             if self.async_llm_client is not None:
-                # Get the most similar program for LLM comparison (thread-safe)
-                loop = asyncio.get_event_loop()
-                most_similar_program = await loop.run_in_executor(
-                    None,
-                    db.get_most_similar_program_thread_safe,
-                    code_embedding,
-                    parent_program.island_idx,
-                )
-
+                # most_similar_program already fetched in the single scan above.
                 if most_similar_program:
                     try:
                         # Read the current proposed code
--- a/shinka/core/novelty_judge.py
+++ b/shinka/core/novelty_judge.py
@@ -85,9 +85,12 @@ class NoveltyJudge:
         }
 
         for attempt in range(self.max_novelty_attempts):
-            # Compute similarities with programs in island
-            similarity_scores = database.compute_similarity(
-                code_embedding, parent_program.island_idx
+            # One island scan yields both the similarity scores and the most
+            # similar program (used below), instead of scanning the island twice.
+            similarity_scores, most_similar_program = (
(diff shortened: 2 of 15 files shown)


RobertTLange:
- #160 database performance
- #161 Slurm polling
- #162 WebUI performance
- #164 LLM client caching

PR #153, opened 2026-07-17
fix: drift bugs from the duplication audit (sync/async twins)

RobertTLange:
## Fix drift bugs from the duplication audit

The duplication audit (`docs/deepreview.md` §4) found ~3,300 duplicated lines dominated by hand-copied **sync↔async twins** (and build↔runtime copies), and — most importantly — **13 latent bugs caused by the copies drifting**, several of them in the copy that actually runs in production while only the (dead) sync copy is tested. This PR fixes those live-path bugs, which the audit lists as the highest-value output and the first consolidation step ("fix the live-path bugs now, independent of refactor"). Each fix ships with a regression test; full suite green; ruff/mypy pass.

**Stacked on `fix/performance`** (base branch of this PR).

### Database / pricing / webui
- **Async event log emitted invalid JSON** — `record_attempt/generation_event_async` used `json.dumps(details)` while the sync log wraps `clean_nan_values(details)`, so NaN/Inf metrics produced non-standard `NaN`/`Infinity` tokens that crash a strict re-parse. Async now matches sync.
- **Weighted parent sampling crashed on a corrupt JSON cell** the DB tolerates elsewhere — its inline parser now defends each field like the canonical `_program_from_row`.
- **Pricing `embedding_overrides` ignored at runtime** — applied only when rebuilding the bundled CSVs, so a live models.dev refresh dropped pinned embedding prices back to upstream. `_apply_overlay` now applies them at runtime too.
- Removed a dead, shadowed `getSelectedNodeId` definition in `viz_tree.html` (the quote-safe `escapeHtml` consolidation already landed in the security PR).

### LLM providers
- **DeepSeek async dropped reasoning-token accounting** — it reported the full `completion_tokens` as output where sync subtracts `thinking_tokens`; both now share one helper.
- **Gemini `thinking_budget` default forked** (1024 sync vs 0 async) — one `DEFAULT_THINKING_BUDGET` constant used by both.
- **Gemini retried an un-retryable error** — the deterministic "does not support structured output" `ValueError` was retried `MAX_TRIES` times; a `giveup=` predicate now surfaces it immediately.
- **Anthropic/DeepSeek aborted a completed generation on an unknown-model pricing miss** — they now fall back to $0 with a warning like OpenAI/local instead of raising.
- **Sync `LLMClient.query` hot-looped on retries** (no sleep) — now sleeps between attempts like its siblings.

### Core async
- **Async summarizer / novelty judge / prompt evolver silently dropped `llm_kwargs`** the sync copies pin, so the live path used default-sampled model/temperature — now forwarded.
- **`_run_fix_patch_async` raised `NameError` on a first `None` LLM response** (swallowed by a broad `except`, discarding costs/metadata) — guarded like its twin `_run_patch_async`.
- **`SystemPromptEvolver.__init__` raised `TypeError` on its documented default** (iterated `patch_types` before the `None`→list normalization) — reordered.
- Deleted a dead method in `async_novelty_judge.py` that called methods defined nowhere (verified zero call sites).

### Structural consolidation — recommended follow-ups (not in this PR)
The audit's consolidation priorities 2–8 (a shared `providers/_retry.py`; a `database/serialization.py` + `schema.py`; collapsing each sync/async twin to one implementation; `plots/style.py`; `shinka-common.css`/`shinka-utils.js`; making `generate_csvs.py` a thin driver over `shinka.pricing`; parametrizing `test_edit_*` and populating `conftest.py`) are larger refactors. The audit notes the async copies currently have **no tests**, so those collapses should land incrementally behind added async-path coverage rather than as one large change here. Fixing the drift bugs first (this PR) removes the immediate correctness risk; the refactors then remove the source of future drift.

### Testing
New regression tests for each drift bug (async event-log NaN, corrupt-JSON parent sampling, runtime embedding overrides, deepseek token accounting, gemini budget/giveup, unknown-model pricing fallback, sync retry sleep, forwarded llm_kwargs, fix-patch first-None guard, evolver default constructor, deleted-method absence). Full suite green; ruff + mypy (CI config) pass.

### DIFF
--- a/shinka/core/async_novelty_judge.py
+++ b/shinka/core/async_novelty_judge.py
@@ -208,6 +208,7 @@ class AsyncNoveltyJudge:
             response = await self.async_llm_client.query(
                 msg=user_msg,
                 system_msg=NOVELTY_SYSTEM_MSG,
+                llm_kwargs=self.async_llm_client.get_kwargs(),
             )
 
             if response is None or response.content is None:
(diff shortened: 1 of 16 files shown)


RobertTLange:
- #163 Gemini retry behavior
- #165 database serialization drift
- #166 provider accounting
- #167 core async parity

PR #154, opened 2026-07-17
refactor: web UI cleanups + deep-review report

RobertTLange:
## Web UI cleanups from deep review

Final category of the deep-review series (`docs/deepreview.md` §5). The Web UI's **security** findings (0.0.0.0 bind, path traversal, stored XSS, missing SRI, PDF JS/file access) landed in the security PR (#151) and its **performance** findings (per-request full-DB reads, uncached summary, expensive poll) in the performance PR (#152). This PR finishes the remaining low-risk maintainability items and documents the large refactors as follow-ups.

**Stacked on `fix/dedup-drift`** (base branch of this PR).

### Done here
- **Consistent quote-safe `escapeHtml`** — `index.html` and `compare.html` now escape quotes too (like the consolidated `viz_tree.html` version from the security PR), so attribute-context interpolations are uniformly safe across all three pages.
- **Dead code removed** — the never-read `WEASYPRINT_AVAILABLE` flag; a duplicated `input:focus-visible` selector in `bandit_selection.css`. (The dead, shadowed `getSelectedNodeId` was removed in the dedup PR #153.)

### Verified NOT dead (left as-is)
While checking the review's dead-code list I found several claims were inaccurate — e.g. the docs demo `throughput` helper is used in 28 places, not dead. Only `formatMoney` is genuinely unused. Given the docs demos have no test harness and negligible product impact, I did not edit them rather than risk breaking a demo on an unverifiable cleanup.

### Structural refactors — recommended follow-ups (not in this PR)
These are real findings but large, and `viz_tree.html` has no JS test harness, so they should land deliberately rather than blind:
- **Split the 18k-line `viz_tree.html` monolith** (D-H1): ~15.9k lines of JS + ~1.5k CSS in one file, with two functions (`renderGraph` ~870 lines, `createCharts` ~1,230 lines) each larger than a typical module. Extract JS into ES modules (data-fetch, tree-render, charts, analysis, prompts) and CSS into a stylesheet shared with `index.html`/`compare.html`, behind a small build step.
- **Extract the duplicated read-DB retry/WAL/close boilerplate** (D-H2): the `max_retries=8` + WAL-PRAGMA + `is_retryable` + `finally: close` block is copy-pasted across ~6 handlers in `visualization.py` (~200 lines). A `with_retrying_readonly_db(...)` helper would remove it — but the handlers now also carry per-handler caching/validation added in the security and performance PRs, so this is best done as its own focused change with the web-UI tests as the safety net.
[...]

PR #158, opened 2026-07-22
fix(webui): sanitize rendered content and PDF export

RobertTLange:
## What changed
- Escape prompt-program table content and remove inline event handlers.
- Sanitize visualization rendering and pin external asset integrity.
- Sandbox wkhtmltopdf and make Pandoc parse LLM content as plain text.
- Keep the Markdown test stub compatible with extension arguments.

## Review scope
Rendered-content extraction from #151 and focused cleanup from #154. Based directly on #150.

## Verification
- Focused XSS, PDF, and visualization tests
- JavaScript syntax/static checks
- Integrated 876-test gate

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -86,7 +86,7 @@ shinka = [
 dev = [
     "pytest>=6.0",
     "pytest-cov",
-    "ruff",
+    "ruff==0.15.20",
     "mypy",
     "black",
     "isort",
--- a/shinka/webui/compare.html
+++ b/shinka/webui/compare.html
@@ -665,9 +665,13 @@
         }
 
         function escapeHtml(text) {
-            const div = document.createElement('div');
-            div.textContent = text;
-            return div.innerHTML;
+            if (text === null || text === undefined) return '';
+            return String(text)
+                .replace(/&/g, "&amp;")
+                .replace(/</g, "&lt;")
+                .replace(/>/g, "&gt;")
+                .replace(/'/g, "&#039;")
+                .replace(/"/g, "&quot;");
         }
 
         // Initialize on page load
--- a/shinka/webui/index.html
+++ b/shinka/webui/index.html
@@ -988,9 +988,13 @@
         }
 
         function escapeHtml(text) {
-            const div = document.createElement('div');
-            div.textContent = text;
-            return div.innerHTML;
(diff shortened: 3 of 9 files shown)


RobertTLange:
## Review follow-up

Resolved each review finding in an isolated commit:

- `eee06a2` — escaped every structured program-summary field before applying the fixed `strong`/`em` Markdown subset, closing the stored-XSS path.
- `207c116` — updated DOMPurify from 3.0.8 to advisory-clean 3.4.12 with an exact, recomputed SHA-384 SRI pin.
- `669bbcc` — sanitized PDF HTML through a zero-attribute tag allowlist and escaped generation metadata; wkhtmltopdf now receives no attacker-controlled resource URLs, with JS/local/external access flags retained as defense in depth.
- `211bb32` — restored the Pandoc fallback with supported HTML input plus `--sandbox`; exhausted conversion now returns an honest HTTP 500 instead of non-PDF bytes under a PDF content type.
- `e4e4203` — removed the collection-global Markdown stub so PDF security tests exercise real Markdown rendering in every test order.
- `629e190` — pinned Ruff 0.15.20 (matching the target branch) after upstream 0.16 rule expansion broke the branch-only push workflow.

Validation:

- Native local gate: Ruff passed; CI-scoped Mypy passed (73 files); pytest + coverage passed (`776 passed, 2 deselected`, 54% overall coverage).
- Focused Web UI/security suite: 19 passed; both cross-module collection orders passed.
- Browser exploit replay: DOMPurify 3.4.12 loaded; zero attacker nodes, event handlers, or script execution; intended strong/emphasis rendering preserved.
- PDF adversarial parser probes: zero disallowed nodes, attributes, or resource requests. Pandoc 3.10.1 HTML/sandbox reader contract verified; a host PDF engine was unavailable locally, and that failure now returns HTTP 500.
- Bandit and `pip-audit`: clean. Final intent, reliability, and security review passes: no remaining material findings.
- GitHub Actions: both push and pull-request `Ruff, Mypy, Pytest` workflows green on `629e190`.

RobertTLange:
Thank you for the rendered-content and PDF hardening work. We are closing the stacked branch while we reset the review base. Please revive this as a small, rebased rendering/PDF-safety PR with focused security tests and the Circle Packing smoke evidence.

PR #160, opened 2026-07-22
perf(db): remove generation hot-path scans

RobertTLange:
## What changed
- Add focused database indexes.
- Replace repeated generation scans and N+1 lookups with bounded queries.
- Preserve existing selection and persistence behavior with regressions.

## Review scope
Database-only performance extraction from #152. Based directly on #150.

## Verification
- 12 focused database tests
- Ruff and Mypy
- Integrated 876-test gate

### DIFF
--- a/shinka/core/async_novelty_judge.py
+++ b/shinka/core/async_novelty_judge.py
@@ -91,9 +91,11 @@ class AsyncNoveltyJudge:
                 return True, novelty_metadata
 
             loop = asyncio.get_event_loop()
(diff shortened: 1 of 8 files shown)


RobertTLange:
Thank you for this runtime/database work. We are avoiding broad async and concurrency changes until they have a Circle Packing matrix at serial and multiple parallelism settings. Please revive only an isolated, rebased change with deterministic regressions and that matrix.

PR #161, opened 2026-07-22
perf(launch): batch Slurm polling

RobertTLange:
## What changed
- Batch Slurm status queries.
- Preserve local-job fallback behavior in mixed batches.
- Keep per-job exceptions in their ordered result slots.
- Cover scheduler fallback ordering and local process failures.

## Review scope
Scheduler-only performance extraction from #152. Based directly on #150.

## Verification
- 13 focused polling tests
- Ruff and Mypy
- Integrated 876-test gate


RobertTLange:
Thank you for the Slurm polling work. We are deferring scheduling/parallelism changes until they have a Circle Packing matrix at serial and multiple parallelism settings. Please revive this as an isolated, rebased change with that evidence.

PR #162, opened 2026-07-22
perf(webui): reduce polling and response work

RobertTLange:
## What changed
- Add cheap change polling and cached summaries.
- Make narrow payloads opt-in and invalidate caches on failed-proposal changes.
- Serve independent requests concurrently.

## Review scope
WebUI-only performance extraction from #152. Review after the server-security replacement.

## Verification
- 24 focused WebUI tests
- Ruff and Mypy
- Integrated 876-test gate

### DIFF
--- a/shinka/webui/visualization.py
+++ b/shinka/webui/visualization.py
@@ -281,6 +281,39 @@ class DatabaseRequestHandler(http.server.SimpleHTTPRequestHandler):
         finally:
             conn.close()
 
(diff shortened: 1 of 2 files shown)


RobertTLange:
Thank you for the WebUI performance work. This branch depends on the current security stack, which we are resetting into smaller PRs. Please revive it after the hardened baseline lands, with a fresh benchmark and smoke evidence.

PR #167, opened 2026-07-22
fix(core): reconcile async behavior

RobertTLange:
## What changed
- Align async core behavior with sync contracts.
- Preserve kwargs, recovery, summarization, and evaluation semantics.
- Add one-concept parity regressions.

## Review scope
Core async-parity extraction from #153. Based directly on #150.

## Verification
- 13 focused core tests
- Ruff and focused Mypy
- Integrated 876-test gate


RobertTLange:
Thank you for this runtime/database work. We are avoiding broad async and concurrency changes until they have a Circle Packing matrix at serial and multiple parallelism settings. Please revive only an isolated, rebased change with deterministic regressions and that matrix.

PR #173, opened 2026-07-26
fix(repo): keep mutation seed trees in the main checkout

anantgar:
## Summary

- Replace three repo-mode seed submodule pointers with regular files.
- Add the seed README and mutable implementation files directly to ShinkaEvolve.
- Keep Headless repo mutations self-contained and available in ordinary checkouts.

## Why

The repo-mode examples pointed at Git submodules. A Headless mutation worktree can therefore receive only a gitlink instead of the seed tree contents, which makes the evaluator-facing repository incomplete and prevents normal file mutation. Keeping the small seed trees in this repository removes that dependency and preserves the existing example contracts.

## Validation

- pytest -q tests/test_repo_agent_evolution.py tests/test_circle_packing_repo_evaluator.py — 20 passed
- git diff --check


anantgar:
Closing this: Shinka does not use repo mode, so this seed-tree change is out of scope.

PR #174, opened 2026-07-28
Add hardened Docker evaluation backend

anantgar:
## Summary

- Add the opt-in `secure_docker` execution backend for single-file candidates.
- Run the unchanged `evaluate.py --program_path --results_dir` interface in a hardened, networkless Docker container.
- Keep host result directories outside the container; a trusted wrapper promotes only bounded JSON results and logs.
- Make CI's established Ruff rule baseline explicit, avoiding unrelated failures after Ruff 0.16 changed its defaults.

## Why

Generated candidate code is untrusted. The existing local backend runs it with the host user's privileges. This backend reduces host exposure while preserving the evaluator interface used by existing tasks.

## Testing

- `uv run ruff check tests --exclude tests/file.py`
- `uv run mypy --follow-imports=skip --ignore-missing-imports tests/test_*.py tests/conftest.py`
- `uv run --with pytest-cov pytest -q -m "not requires_secrets and not models_dev_live" --cov=shinka --cov-report=term-missing --cov-report=xml:coverage.xml`
- `python3 -m pytest tests/test_secure_docker.py tests/test_scheduler_env_activation.py tests/test_shinka_run_cli.py -q`
- `uv run --group docs mkdocs build --strict`

All checks pass. The coverage run reports 751 passed and 2 deselected; focused backend tests report 59 passed.

## Risks and compatibility

- Requires a locally available, digest-pinned evaluator image and Docker's default seccomp profile; Linux requires rootless Docker unless explicitly configured for a dedicated VM.
- Docker is host containment, not a complete security boundary: the daemon, image, kernel, and runtime remain trusted.
- Candidate and evaluator intentionally share one container for compatibility. This does not prevent evaluator inspection, reward hacking, or score forgery by an adversarial candidate.
- Only `metrics.json`, `correct.json`, and bounded logs return to the host. Other evaluator artifacts remain in the container.
- No default local or SLURM backend behavior changes.

## Evaluation evidence

- An end-to-end `secure_docker.submit(...)` smoke test used `python@sha256:57cd7c3a7a273101a6485ba99423ee568157882804b1124b4dd04266317710de`.
- The test evaluator invoked a one-file Python candidate through the ordinary `--program_path` and `--results_dir` arguments. The candidate printed `42`; the evaluator emitted `{"score": 1.0}` and `{"correct": true}`.
- The container exited with status 0, the host received only the expected JSON result files, and the managed container was automatically removed.
[...]

PR #175, opened 2026-07-31
perf(database): parse each program's AST once in analyze_code_metrics

dexhunter:
## Summary

- Parse valid Python source once during complexity analysis and reuse the AST for Radon cyclomatic, Halstead, and nesting metrics.
- Preserve `max_nesting_depth(code_string)` and every returned metric key/value.
- Add exact-contract regressions for Python, invalid-Python fallback, Go, and single-parse behavior.
- Credit @dexhunter in the changelog.

## Verification

- `uv run pytest -q tests/test_complexity.py` — 4 passed.
- Database-adjacent tests — 16 passed.
- Ruff and Mypy focused checks — clean.
- Full CI-equivalent gate — 731 passed, 7 skipped, 2 deselected.
- `shinka/database/complexity.py` coverage — 93%.
- Fresh GitHub CI — green on `be04d05`.

## Compatibility and rollback

Public call signatures and Python/non-Python metric contracts are unchanged. Revert `be04d05` to restore the prior three-parse implementation.

## Weco dashboard

[View the optimization run](https://dashboard.weco.ai/share/XpTb21O9z1DcWjRs9EFjbggGSWVr6QFl)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,8 @@ All notable changes to `shinka-evolve` are documented in this file.
 
 ### Changed
 
+- Reduced Python complexity-analysis overhead by parsing each candidate AST once
+  while preserving the existing metrics contract. Thanks @dexhunter.
 - Relaxed the exact HTTPX dependency pin to a `>=0.27` compatibility lower
   bound in issue #180. Thanks @htmai-880.
 - Removed the automatic Claude Code Review pull-request workflow.
--- a/shinka/database/complexity.py
+++ b/shinka/database/complexity.py
@@ -1,43 +1,43 @@
 import ast
-from radon.complexity import cc_visit
-from radon.metrics import h_visit
-from radon.raw import analyze
 import math
 import re
+
+from radon.complexity import cc_visit_ast
+from radon.metrics import h_visit_ast
+from radon.raw import analyze
+
+
+class _NestingVisitor(ast.NodeVisitor):
+    """Track the deepest chain of nested control structures."""
+
+    def __init__(self):
+        self.current_depth = 0
+        self.max_depth = 0
+
+    def _visit_nested(self, node):
+        self.current_depth += 1
+        self.max_depth = max(self.max_depth, self.current_depth)
+        self.generic_visit(node)
+        self.current_depth -= 1
+
+    visit_If = _visit_nested
+    visit_For = _visit_nested
+    visit_While = _visit_nested
+    visit_With = _visit_nested
+    visit_Try = _visit_nested
+    visit_FunctionDef = _visit_nested
+    visit_AsyncFunctionDef = _visit_nested
+
+
+def _max_nesting_depth(tree):
+    visitor = _NestingVisitor()
+    visitor.visit(tree)
+    return visitor.max_depth
 
 
 def max_nesting_depth(code_string):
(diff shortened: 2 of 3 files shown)


dexhunter:
Heads-up on the red check: the `Ruff` step is failing for a reason unrelated to this PR,
and it will affect any PR whose CI runs from now on.

`uv.lock` is gitignored, so `uv sync --dev` resolves the newest `ruff` at job time, and
`[dependency-groups] dev` lists a bare `ruff` with no constraint. **ruff 0.16.1 was
released 2026-07-30**, promoting rules (`UP006`, `UP045`, …) that flag 153 pre-existing
issues under `tests/`.

Reproduced on unmodified `main`, zero diff:

```
$ git clone --depth 1 --branch main https://github.com/SakanaAI/ShinkaEvolve.git && cd ShinkaEvolve
$ git rev-parse HEAD
b67a07328ab7e21e999d9e20a44f4f0054a4b83c
$ git status --porcelain      # empty
$ uv sync --dev && uv run ruff --version
ruff 0.16.1
$ uv run ruff check tests --exclude tests/file.py
Found 153 errors.
$ echo $?
1
```

The last green `Ruff, Mypy, Pytest` on `main` completed 2026-07-16T11:32:12Z, before that
ruff release.

This PR changes one file, `shinka/database/complexity.py`, and touches nothing under
`tests/`. On the same ruff 0.16.1 that file reports the same two findings (`I001`,
`PIE810`) before and after the change, so it adds none.

I have deliberately not bundled a fix, since the remedy is your call — constraining ruff
in `[dependency-groups]`, committing `uv.lock`, or landing a `ruff check --fix` sweep over
`tests/` would each be a much larger and unrelated diff. Happy to open a separate PR for
whichever you prefer.

dexhunter:
Following up on my own comment: I went looking to see whether a standalone fix was worth
opening, and it turns out you already have two, both from @anantgar, and both already open.

- #171 pins the linter: `"ruff"` becomes `"ruff>=0.15.21,<0.16"`
- #174 makes the rule set explicit: `[tool.ruff.lint] select = ["E4", "E7", "E9", "F"]`

That also explains why those branches are green while this one is red — each carries its
own remedy, so it was never a question of when their CI happened to run.

I checked both against pristine `main` in case it helps you pick:

| approach | result |
|---|---|
| #174's explicit `select`, on current ruff **0.16.1** | `All checks passed!` |
| #171's `<0.16` pin (resolves 0.15.22) | passes |
| unmodified `main`, ruff 0.16.1 | 153 errors |

Both work. The difference is that #174's version survives the next ruff release, because it
states the rule set instead of freezing the tool — the 153 findings are almost entirely new
default rules in 0.16.0 (86 of them are `RUF059` alone), not new problems in your code. The
pin fixes today and defers the same break to whenever someone bumps it.

One thing you may or may not care about: right now the fix only exists inside two feature
PRs, so getting CI green means merging either a Gemini provider change or a 2,240-line
Docker backend. I was going to offer a one-line standalone PR, but it would just be
@anantgar's change retyped, and it's their fix — so I'll leave it with you rather than
duplicate it. Happy to open one if you'd rather have it separated out.

Nothing here needs anything from this PR; it's unblocked the moment either of those lands.

anantgar:
Feel free to open the 1 line PR!

PR #185, opened 2026-08-16
fix(edit): keep EVOLVE-BLOCK-END marker on its own line for insertions

### LINKED ISSUE #183: [Bug]: insertion diffs glue the payload onto the EVOLVE-BLOCK-END marker (compounding corruption; all insertions rejected for block-comment languages)
### Summary

Insertion diffs (empty `SEARCH` blocks) splice the payload directly against the `EVOLVE-BLOCK-END` marker. `apply_search_replace` strips the payload's trailing newline (`_strip_trailing_whitespace`) and then inserts at the END marker's match position (`shinka/edit/apply_diff.py`, insertion branch of `apply_search_replace`), so the marker is glued onto the payload's last line.

Consequences by language:

1. **Python / C-style comment languages:** the patch "succeeds" (`num_applied=1`, no error) but the program now ends with `y = 2# EVOLVE-BLOCK-END`. The corruption compounds: a second insertion on that program splices into the middle of the glued line and produces `y = 2z = 3# EVOLVE-BLOCK-END`, which is a syntax error. The evaluation failure is then attributed to the LLM proposal rather than the patch applier.
2. **Block-comment languages (Wolfram; same class as HTML/Markdown markers):** the post-apply marker validation correctly notices the malformed marker line and rejects the patch, so **every** insertion-style diff fails for these languages, with an error that blames the model output: `Line 3: malformed EVOLVE-BLOCK-END marker. Got: 'g[x_] := x^3(* EVOLVE-BLOCK-END *)'`.

This still reproduces on the `fix/correctness-concurrency` deep-review branch (PR #150), so it does not overlap with the pending split PRs.

### Steps to reproduce

```python
from shinka.edit.apply_diff import apply_diff_patch
import ast

original = "# EVOLVE-BLOCK-START\ndef f(x):\n    return x\n# EVOLVE-BLOCK-END\n"
ins = lambda payload: f"<<<<<<< SEARCH\n\n=======\n{payload}\n>>>>>>> REPLACE\n"

u1, n1, *_ = apply_diff_patch(ins("y = 2"), original_str=original, language="python", verbose=False)
print(n1)          # 1
print(u1.splitlines()[-1])  # 'y = 2# EVOLVE-BLOCK-END'  <- marker glued

u2, n2, _, err, *_ = apply_diff_patch(ins("z = 3"), original_str=u1, language="python", verbose=False)
print(n2, err)     # 1 None
ast.parse(u2)      # SyntaxError: invalid decimal literal ('y = 2z = 3# EVOLVE-BLOCK-END')
```

Wolfram variant (every insertion rejected):

```python
wl = "(* EVOLVE-BLOCK-START *)\nf[x_] := x\n(* EVOLVE-BLOCK-END *)\n"
u, n, _, err, *_ = apply_diff_patch(ins("g[x_] := x^3"), original_str=wl, language="wolfram", verbose=False)
print(n)    # 0
print(err)  # EVOLVE-BLOCK marker validation failed ... Got: 'g[x_] := x^3(* EVOLVE-BLOCK-END *)'
```

### Expected behavior

The inserted payload ends with a newline and the `EVOLVE-BLOCK-END` marker keeps its own line (and indentation), e.g.

```
def f(x):
    return x
y = 2
# EVOLVE-BLOCK-END
```

Insertion patches for Wolfram and other block-comment languages apply cleanly.

### Actual behavior

```
def f(x):
    return x
y = 2# EVOLVE-BLOCK-END
```

and after a second insertion:

```
y = 2z = 3# EVOLVE-BLOCK-END   # SyntaxError at evaluation time
```

For Wolfram, `num_applied=0` with a marker-validation error on every insertion diff.

### Environment

Python 3.12, macOS, `uv sync --dev`, current `main` (2bf8cfe). No provider credentials needed; pure string-level reproduction.

### Logs and artifacts

Root cause: in `apply_search_replace`, the insertion branch does `new_text[:b] + replace + new_text[b:]` where `b` is `EVOLVE_END` match start and `replace` has been stripped of its trailing newline, so nothing separates the payload from the marker. Fix: insert at the start of the marker's line and restore the payload's trailing newline. PR with regression tests incoming.

Atharva-Kanherkar:
Fixes #183.

Empty-SEARCH insertion diffs spliced the payload at the EVOLVE-BLOCK-END marker's position after the payload's trailing newline was stripped, gluing the marker onto the last inserted line. Consecutive insertions then merged code lines into invalid programs, and block-comment languages (Wolfram) rejected every insertion at marker validation.

Fix: insert at the start of the marker's line and restore the payload's trailing newline, so the marker keeps its own line and indentation.

Tests: four regressions in `tests/test_edit_base.py`, all failing without the fix. Full suite green (823 passed), ruff and mypy clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,8 @@ All notable changes to `shinka-evolve` are documented in this file.
   against the EVOLVE-BLOCK-END marker, which corrupted the marker line, let
   consecutive insertions merge code lines into invalid programs, and caused
   marker validation to reject every insertion patch for block-comment
-  languages such as Wolfram. Reported in issue #183.
+  languages such as Wolfram. Reported in issue #183. Thanks
+  @Atharva-Kanherkar.
 - Fixed Gemini retry handling so unsupported structured-output requests fail
   immediately, synchronous retries are paced, and sync/async calls share the
   same default thinking budget.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,11 @@ All notable changes to `shinka-evolve` are documented in this file.
 
 ### Fixed
 
+- Fixed diff insertions (empty SEARCH blocks) splicing the payload directly
+  against the EVOLVE-BLOCK-END marker, which corrupted the marker line, let
+  consecutive insertions merge code lines into invalid programs, and caused
+  marker validation to reject every insertion patch for block-comment
+  languages such as Wolfram. Reported in issue #183.
 - Fixed Gemini retry handling so unsupported structured-output requests fail
   immediately, synchronous retries are paced, and sync/async calls share the
   same default thinking budget.
--- a/shinka/edit/apply_diff.py
+++ b/shinka/edit/apply_diff.py
@@ -634,8 +634,20 @@ def apply_search_replace(
             # of that line; when code precedes the marker on the same line,
             # append after that code and move the marker to its own line.
             line_start = new_text.rfind("\n", 0, b) + 1
-            if new_text[line_start:b].strip():
-                new_text = new_text[:b] + "\n" + replace + "\n" + new_text[b:]
+            marker_line_prefix = new_text[line_start:b]
+            if marker_line_prefix.strip():
+                leading_whitespace = marker_line_prefix[
+                    : len(marker_line_prefix) - len(marker_line_prefix.lstrip())
+                ]
+                new_text = (
+                    new_text[:line_start]
+                    + marker_line_prefix.rstrip()
+                    + "\n"
+                    + replace
+                    + "\n"
+                    + leading_whitespace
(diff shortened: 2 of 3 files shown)


RobertTLange:
Maintainer follow-up landed in `04ecab6`:

- preserve indentation when recovering an inline `EVOLVE-BLOCK-END` marker
- remove the obsolete inline-comment separator whitespace
- strengthen the same-line regression with exact nested output
- credit @Atharva-Kanherkar in `CHANGELOG.md`

Validation: Ruff clean; mypy clean across 70 files; 818 passed, 7 skipped, 1 deselected. No review threads were present to resolve. Residual risk is limited to the existing marker parser behavior.

Thanks @Atharva-Kanherkar for the report and contribution.

PR #188, opened 2026-08-20
fix(llm): key anthropic response parsing on block type

### LINKED ISSUE #187: [Bug]: anthropic provider extracts response content by position, crashing on redacted_thinking and thinking-only responses
### Summary

`query_anthropic` and `query_anthropic_async` separate the thinking and text parts of a response by positional index instead of by block type (`shinka/llm/providers/anthropic.py`, lines 94-100 for the sync path and 172-178 for the async path):

```python
# Separate thinking from non-thinking content
if len(response.content) == 1:
    thought = ""
    content = response.content[0].text
else:
    thought = response.content[0].thinking
    content = response.content[1].text
```

`response.content` is a list of typed blocks (`text`, `thinking`, `redacted_thinking`, and others), and each type carries different attributes. This code assumes every multi-block response is exactly `[thinking, text]`, so it reads `.thinking` off block 0 and `.text` off block 1 unconditionally.

This is a live path rather than a hypothetical one. Extended thinking is enabled for Anthropic and Bedrock reasoning models in `shinka/llm/kwargs.py` (`kwargs_dict["thinking"] = {"type": "enabled", "budget_tokens": thinking_tokens}`), and models that report `requires_reasoning` have the effort forced up from `disabled` to `low`, so thinking blocks are expected in normal operation.

Three failures follow:

1. **`redacted_thinking` blocks.** When a thinking block is flagged and returned encrypted, it carries `data` and has neither `thinking` nor `text`. Both `[thinking, redacted_thinking, text]` and `[redacted_thinking, text]` raise `AttributeError`.
2. **Thinking-only responses.** If generation stops while still inside the thinking block, `content` holds a single thinking block, the `len(...) == 1` branch reads `.text` off it, and it raises `AttributeError`.
3. **Silent content loss on three or more blocks.** `[thinking, text, text]` does not crash. It returns block 1 and discards every later text block, so the recorded answer is quietly truncated with no error.

The crashes cost more than an ordinary exception would, because the extraction runs inside the `backoff.on_exception` decorator whose retry tuple is `(APIConnectionError, APIStatusError, RateLimitError, APITimeoutError)`. `AttributeError` is not in that tuple, so the query aborts on the first occurrence and the generation is lost instead of retried.

The multi-block `else` branch currently has no test coverage. `tests/test_anthropic_accounting.py` is the only Anthropic test, and every response it constructs has a single `SimpleNamespace(text="ok")` block, which only ever exercises the `len(...) == 1` branch.

### Steps to reproduce

Save as `repro.py` in the repository root and run `uv run python repro.py`. No provider credentials are needed; the client is stubbed the same way `tests/test_anthropic_accounting.py` stubs it.

```python
from types import SimpleNamespace
from shinka.llm.providers.anthropic import query_anthropic

class _Client:
    def __init__(self, response):
        self.messages = SimpleNamespace(create=lambda **_kw: response)

def _response(blocks):
    return SimpleNamespace(
        content=[SimpleNamespace(type=t, **a) for t, a in blocks],
        usage=SimpleNamespace(input_tokens=10, output_tokens=20, output_tokens_details=None),
    )

CASES = [
    ("thinking + text (intended shape)", [("thinking", {"thinking": "t"}), ("text", {"text": "answer"})]),
    ("thinking + redacted_thinking + text", [("thinking", {"thinking": "t"}), ("redacted_thinking", {"data": "enc"}), ("text", {"text": "answer"})]),
    ("single thinking block", [("thinking", {"thinking": "t"})]),
    ("redacted_thinking + text", [("redacted_thinking", {"data": "enc"}), ("text", {"text": "answer"})]),
    ("thinking + text + text", [("thinking", {"thinking": "t"}), ("text", {"text": "part one"}), ("text", {"text": "part two"})]),
]

for label, blocks in CASES:
    try:
        r = query_anthropic(_Client(_response(blocks)), "claude-opus-4-8", "msg", "sys", [], None)
        print(f"  ok    {label}: content={r.content!r}")
    except Exception as e:
        print(f"  crash {label}: {type(e).__name__}: {e}")
```

The async path fails identically, since `query_anthropic_async` carries the same six lines.

### Expected behavior

Content and thinking are collected by inspecting each block's `type`, so any ordering or combination of block types is handled. Specifically:

- `[thinking, redacted_thinking, text]` returns `content="answer"` and does not raise.
- `[redacted_thinking, text]` returns `content="answer"` and does not raise.
- A thinking-only response returns empty content rather than raising.
- `[thinking, text, text]` returns both text segments joined, not just the first.

### Actual behavior

```
  ok    thinking + text (intended shape): content='answer'
  crash thinking + redacted_thinking + text: AttributeError: 'types.SimpleNamespace' object has no attribute 'text'
  crash single thinking block: AttributeError: 'types.SimpleNamespace' object has no attribute 'text'
  crash redacted_thinking + text: AttributeError: 'types.SimpleNamespace' object has no attribute 'thinking'
  ok    thinking + text + text: content='part one'
```

The last line is the silent case: `part two` is dropped with no error raised.

### Environment

Python 3.13.14, macOS 26.2, `uv sync --dev`, current `main` (e344678). `anthropic` 0.121.0 (`pyproject.toml` leaves it unpinned). No provider credentials needed; the reproduction is a pure stub-level test of the extraction branch.

### Possible fix

Dispatch on block type instead of position, in both `query_anthropic` and `query_anthropic_async`:

```python
thought_parts, text_parts = [], []
for block in response.content:
    btype = getattr(block, "type", None)
    if btype == "text":
        text_parts.append(block.text)
    elif btype == "thinking":
        thought_parts.append(block.thinking)
    elif btype == "redacted_thinking":
        thought_parts.append(getattr(block, "data", ""))
content = "\n".join(text_parts)
thought = "\n".join(thought_parts)
```

This covers all four failing shapes above and keeps the intended `[thinking, text]` result unchanged. Unknown or future block types are skipped rather than crashing. A regression test over these shapes would fit alongside the existing cases in `tests/test_anthropic_accounting.py`.

### Checks

- [x] I searched existing issues and did not find a duplicate.
- [x] I can reproduce this on the current main branch or current release.

Atharva-Kanherkar:
## Summary

- `query_anthropic` and `query_anthropic_async` derived `content` and `thought` from positional indices (`response.content[0].thinking`, `response.content[1].text`), which assumes every multi-block response is exactly `[thinking, text]`. Both call sites now go through one `split_content_blocks` helper that accumulates parts by inspecting each block's `type`.
- Added regression coverage for the block shapes that previously raised or truncated, run against both query paths.

## Why

`response.content` is a list of typed blocks whose attributes differ per type, and neither the ordering nor the count is guaranteed. Extended thinking is enabled for Anthropic and Bedrock reasoning models in `shinka/llm/kwargs.py`, and models reporting `requires_reasoning` have the effort forced up from `disabled` to `low`, so multi-block responses occur in normal operation.

Three concrete failures:

1. `redacted_thinking` blocks carry `data` and have neither `thinking` nor `text`, so `[thinking, redacted_thinking, text]` and `[redacted_thinking, text]` both raised `AttributeError`.
2. A response that stops inside the thinking block leaves a single thinking block, so the `len(...) == 1` branch read `.text` off it and raised `AttributeError`.
3. `[thinking, text, text]` did not raise; it returned block 1 and discarded every later text block, so the recorded answer was silently truncated.

The crashes were worse than an ordinary exception because extraction runs inside the `backoff.on_exception` decorator, whose retry tuple is `(APIConnectionError, APIStatusError, RateLimitError, APITimeoutError)`. `AttributeError` is not in it, so the query aborted on first occurrence and the generation was lost rather than retried.

The multi-block branch had no test coverage: every response built in `tests/test_anthropic_accounting.py` had a single text block, exercising only the `len(...) == 1` path.

## Linked issue or context

Fixes #187.

## Testing

- Commands run:
  - `uv run --with pytest-cov pytest -q -m "not requires_secrets and not models_dev_live" --cov=shinka --cov-report=term-missing`
- Optional secret-backed tests: not run, no provider credentials configured in this environment. The change is confined to parsing an already-returned response object, and the added tests stub the client the same way the existing tests in that module do.
- Results:
  - `ruff check`: all checks passed. `ruff format --check` also clean on both touched files.
  - `mypy`: success, no issues found in 75 source files.
  - `pytest`: 894 passed, 2 deselected.
  - `tests/test_anthropic_accounting.py`: 18 passed.
  - Confirmed the new tests are meaningful rather than tautological: with the provider reverted to its previous version, 10 of them fail (the 5 broken shapes across both paths) while the 2 already-working shapes and the 4 pre-existing accounting tests still pass. With the fix, all 18 pass.

## Risks and compatibility

- No API or signature change. `split_content_blocks` is additive and both existing call sites keep returning the same `(content, thought)` pair.
- The intended `[thinking, text]` shape returns byte-identical output to before, so normal runs are unaffected.
- Multiple text blocks now join with a newline instead of silently dropping all but the first. That is a behaviour change, in the direction of not losing model output.
- Block types the helper does not recognise, including `tool_use` and any future type, are skipped rather than raising. This repository does not currently send `tools`, so no `tool_use` block is expected today; skipping is defensive so a later change cannot reintroduce the crash.
- Cost accounting (`get_anthropic_costs`), the `new_msg_history` assistant entry, and the structured-output branch are untouched.

Not applicable. This does not touch parent selection, mutation or edit generation, archive updates, prompt evolution, novelty logic, evaluation scheduling, proposal oversubscription, or island behaviour. It is confined to parsing one provider's response payload.

## Docs and UI

- No docs change needed; no user-facing configuration, CLI output, or WebUI behaviour changes.
- No UI change, so no screenshots.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,9 @@ All notable changes to `shinka-evolve` are documented in this file.
 
 ### Fixed
 
+- Fixed Anthropic response parsing to dispatch by content-block type, avoiding
+  crashes and truncated output for redacted, thinking-only, and multi-block
+  responses in PR #188. Thanks @Atharva-Kanherkar.
 - Fixed Anthropic and DeepSeek token/cost accounting parity across synchronous
   and asynchronous queries, including thinking-token decomposition and safe
   zero-cost fallback for models missing from the pricing catalog. Runtime
--- a/shinka/llm/providers/anthropic.py
+++ b/shinka/llm/providers/anthropic.py
@@ -41,6 +41,26 @@ def get_anthropic_costs(response, model):
     }
 
 
+def split_content_blocks(response):
+    """Split response blocks into visible text and thinking, keyed on block type.
+
+    Anthropic returns a list of typed blocks (text, thinking, redacted_thinking,
+    tool_use, ...) whose attributes differ per type, and the ordering and count
+    are not guaranteed. Unknown types are skipped.
+    """
+    text_parts = []
+    thought_parts = []
+    for block in response.content:
+        block_type = getattr(block, "type", None)
+        if block_type == "text":
+            text_parts.append(block.text)
+        elif block_type == "thinking":
+            thought_parts.append(block.thinking)
+        elif block_type == "redacted_thinking":
+            thought_parts.append(getattr(block, "data", ""))
+    return "\n".join(text_parts), "\n".join(thought_parts)
+
+
 def backoff_handler(details):
     exc = details.get("exception")
     if exc:
@@ -91,13 +111,7 @@ def query_anthropic(
             messages=new_msg_history,
             **kwargs,
         )
-        # Separate thinking from non-thinking content
-        if len(response.content) == 1:
-            thought = ""
-            content = response.content[0].text
-        else:
-            thought = response.content[0].thinking
-            content = response.content[1].text
+        content, thought = split_content_blocks(response)
     else:
         raise NotImplementedError("Structured output not supported for Anthropic.")
     new_msg_history.append(
@@ -169,13 +183,7 @@ async def query_anthropic_async(
             messages=new_msg_history,
             **kwargs,
         )
-        # Separate thinking from non-thinking content
-        if len(response.content) == 1:
-            thought = ""
-            content = response.content[0].text
-        else:
-            thought = response.content[0].thinking
-            content = response.content[1].text
+        content, thought = split_content_blocks(response)
     else:
         raise NotImplementedError("Structured output not supported for Anthropic.")
     new_msg_history.append(
--- a/tests/test_anthropic_accounting.py
+++ b/tests/test_anthropic_accounting.py
@@ -1,7 +1,9 @@
-"""Regression tests for Anthropic pricing fallback parity."""
+"""Regression tests for Anthropic pricing parity and response block parsing."""
 
 import asyncio
 from types import SimpleNamespace
+
+import pytest
 
 import shinka.llm.providers.anthropic as anthropic_provider
 from shinka.llm.providers.anthropic import (
@@ -22,7 +24,7 @@ def _response(
         else SimpleNamespace(thinking_tokens=thinking_tokens)
     )
     return SimpleNamespace(
-        content=[SimpleNamespace(text="ok")],
+        content=[SimpleNamespace(type="text", text="ok")],
         usage=SimpleNamespace(
             input_tokens=input_tokens,
             output_tokens=output_tokens,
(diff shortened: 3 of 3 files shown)


RobertTLange:
Merged—thank you @Atharva-Kanherkar for tracking down the Anthropic content-block parsing failures and adding thorough sync/async regression coverage! Your contribution is credited in the changelog as part of the merge.

